PR #31, opened 2024-02-18
add parsing of cik_lookup_data file url embedded here

rla3rd:
https://www.sec.gov/os/accessing-edgar-data
https://www.sec.gov/Archives/edgar/cik-lookup-data.txt

### DIFF
--- a/edgar/_companies.py
+++ b/edgar/_companies.py
@@ -40,7 +40,8 @@ __all__ = [
     'get_company_tickers',
     'get_entity_submissions',
     'parse_entity_submissions',
-    'get_ticker_to_cik_lookup'
+    'get_ticker_to_cik_lookup',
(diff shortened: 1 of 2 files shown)


dgunning:
Thanks, I will take a look.
If this adds Person lookup then it's functionality that's needed .. there's already company lookup

dgunning (commented):
This contribution is much appreciated. This opens up enhancements to Person lookup and display.

My plan is to accept and test it, though I might make a change to use a common download_text() function in place of get.

Note that some of the data retrieval apis may change in the future as part of a performance enhancement release. 

Thanks

dgunning (approved):
Approved

dgunning:
Merge cis_lookup contribution

Issue #34, opened 2024-03-11
Get latest press release per company / query by attachment name and date


paulommaia:
I think press releases are typically encoded as EX 99.1, which is an attachment of 8-K. This worked for me in ~200 out of 500 companies, by iterating through every 8-K and checking for that - but the other ones were missing

Is there an easier way to get the latest press release per company?

There are functions to get the latest 8-K but this doesn't always have the correct attachment

dgunning:
Thanks for the suggestion. I will look into it, maybe add press release as a property of 8-Ks

dgunning:
Not all 8-K's have a Press Release attached, as your numbers show. I have added a property called press_release  that will return the press release if it has one

```
filings = get_filings(form='8-K')
filiing = filings[0]
eightk = filing.obj()
press_release = eight.press_release
 
```

dgunning:
I have completed and tested this feature. Will release tomorrow

dgunning:
Released version 2.11.0

paulommaia:
Thanks, I've actually found that some companies name the press release as EX 99 and not EX 99.1, see for example ACN:

`
cik = 'ACN'
after = '2023-01-01'
filings = Company(cik).get_filings(form="8-K").filter(date=f"{after}:")
`

NEWS RELEASE OF ACCENTURE, DATED DECEMBER 19, 2023     fy24q1earnings8-kexhibit.htm   EX-99        324.4 KB

dgunning:
OK. Good catch. Will look into this.

dgunning:
There are many different ways to label Exhibit 99. 
<img width="295" alt="ExhibitCounts" src="https://github.com/dgunning/edgartools/assets/1523271/735c83be-bbf5-4199-9c39-b15c9930c719">

dgunning:
<img width="767" alt="PressRelease" src="https://github.com/dgunning/edgartools/assets/1523271/aa555cb3-649a-409c-b30b-aacc6fb96ddc">

dgunning:
Released 2.13.0

```python
eightk = filing.obj()

# if press release present
press_releases:PressReleases = eightk.press_releases
press_release:PressRelease = press_releases[0]
```

dgunning:
Fixed in 2.13.0

PR #38, opened 2024-03-27
Added support for press releases in EX-99 attachments.

vgreg:
A fix for Issue https://github.com/dgunning/edgartools/issues/34

Added `EX-99` as a second type of valid attachment for press releases.

All tests are passing, except for `test_form4_to_markdown()` which fails on line 35, but that is unrelated to my changes.

### DIFF
--- a/edgar/company_reports.py
+++ b/edgar/company_reports.py
@@ -491,14 +491,14 @@ class EightK():
     @property
     def has_press_release(self):
         attachments: Attachments = self._filing.attachments
-        # press release is Type EX-99.1
-        return attachments.query("Type=='EX-99.1'") is not None
+        # press release is Type EX-99 or EX-99.1
+        return attachments.query("Type=='EX-99' | Type=='EX-99.1'") is not None
 
     @property
     def press_release(self):
         attachments: Attachments = self._filing.attachments
-        # press release is Type EX-99.1
-        press_release_results = attachments.query("Type=='EX-99.1' & Document.str.contains('htm')")
+        # press release is Type EX-99 or EX-99.1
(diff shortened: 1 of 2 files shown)


dgunning:
Are press releases just EX-99 and EX-99.1 or EX-99.2?

In a sample of 500 filings there are multiple ways to write EX-99.1 e.g. EX-99.01. 
ExhibitCounts

Is the intent of this press release functionality "Show me the attachments that by convention are intended to be press releases" ?

dgunning:
PressRelease

vgreg:
That would be my use case. I will add the other types as well, or maybe we would be better off with a wildcard like "EX-99*"?

Another thing your example made me realize is that currently, the code for press releases returns only the first one. But in your example, I would want to get all the press releases. Is it worth building this into the library or should I access the attachments directly in my code?

dgunning:
The logic I have so far is

`Description.str.match('.*RELEASE') or Type.isin(['EX-99.1', 'EX-99', 'EX-99.01'])`. This will include EX-99.2+ marked as press releases. Sometimes they are named 'NEWS RELEASE' 

In addition
`Document.str.endswith('.htm')` to select only the HTML versions. 

I am also changing the variable to handle multiple `press_releases`.

vgreg:
That should do it. Thanks!

dgunning:
Released 2.13.0

```python
eightk = filing.obj()

# if press release present
press_releases:PressReleases = eightk.press_releases
press_release:PressRelease = press_releases[0]
```

dgunning:
Added equivalent functionality in 2.13.0

PR #123, opened 2024-10-07
Add a fallback for the CIK lookup.

MauritsBrinkman:
I have an application running in production and today I experienced an issue when there was an outage of the edgar web page that is used by the CIK lookup. Although edgartools uses the `@lru_cache` decorator, this didn't prevent the `Company` object from giving an error because it couldn't use the CIK lookup. Luckily, this case might be solved easily by having a fallback that is dependent on [this JSON file provided by SEC](https://www.sec.gov/files/company_tickers.json).

The lookup is dependent on ticker.txt: tab-delimited text file (ticker.txt) for ticker / CIK. This is fragile, as it might be (although we have caching) that an outage of this source results in failure of the Company(<ticker>) object. Now, having company_tickers.json as a second fallback for the JSON object, an outage of the ticker.txt will not necessarily result in an error for the use of the Company(<ticker>).

### DIFF
--- a/edgar/reference/tickers.py
+++ b/edgar/reference/tickers.py
@@ -1,4 +1,5 @@
 import re
+import json
 from functools import lru_cache
 from io import StringIO
 from typing import Optional, Union, List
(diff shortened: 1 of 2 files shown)


dgunning:
Was this due to the SEC outage this afternoon? Could the outage and future outage affect both cik lookups?

I'm also thinking of making more of edgartools use local storage, including the cik lookup

dgunning (approved):
Looks good. The fallback adds a little redundancy though it's possible that both urls might be down at the same time.

I will approve test and integrate

PR #134, opened 2024-10-26
Minor refactor of regex patterns to use raw strings

webbsledge:
Updated regex patterns in documents.py, funds.py, and htmltools.py to use raw strings with an 'r' prefix to prevent escape sequence issues.


dgunning (approved):
Reviewed. Looks good. Thanks

PR #141, opened 2024-11-02
Added function to find individual company ticker from CIK value

webbsledge:
Implement simple function to lookup a ticker for a single CIK value. I didn't see this implemented anywhere already.


dgunning (commented):
This is probably needed so thanks for this.

2 issues - a cik might have more than one ticker - do you intend to return just the first one?

We also need to handle the case when the ticker is not found so that it does not result in an exception

webbsledge:
Made some changes to return a list of tickers or a single ticker, depending. Also added some exception handling... hopefully not overkill. Thanks for your patience with me. This is my third ever PR. 😆

dgunning (approved):
Looks good. Will integrate and test

PR #147, opened 2024-11-09
Minor cleanup in FinancialStatements.ipynb

webbsledge:
Potentially remove line in notebook. `xb` is not defined in the notebook, and I can't tell what the intent with this line is or was. **Please reject this commit if I am missing something,** but at the moment it's confusing me. Referencing the `financials` object directly seems to work as intended.


dgunning (approved):
Approved

PR #167, opened 2024-12-19
fix IPython RuntimeError for nest_asyncio

rla3rd:
ipython had a runtime error rather than starting nest_asyncio gracefully, this bit of code fixes it


rla3rd:
damn, ok so this fixes if using the ipython environment natively, it breaks jupyter.

PR #170, opened 2024-12-22
Dev

rla3rd:
make edgar_mode and env variable and improve the _from_local function calls


dgunning (commented):
So this adds the ability to set the access mode via an environmental variable

dgunning (approved):
Reviewed and approved. Will integrate and test

PR #173, opened 2024-12-24
fix forms page missing GET variable

rla3rd:
forms fetcher missing the page GET variable


dgunning (approved):
Thanks for this change. Accepting and testing

PR #176, opened 2024-12-28
fix: allow for non breaking spaces in item pattern

raminos:
Hi!

I came across an edge case where the library was not correctly detecting the first item of a 10-K (the latest 10-K of LMT). The issue is that the 10-K has a non breaking space after the word 'Item'.

Here the HTML from the filing:

```html
<span style="color:#000000;font-family:\'Times New Roman\',sans-serif;font-size:10pt;font-weight:700;line-height:120%">ITEM&#160; 1.</span>
```
I've adjusted the pattern to allow for non breaking spaces and to allow a few spaces. Now on my side the Item gets detected.

### DIFF
--- a/edgar/files/htmltools.py
+++ b/edgar/files/htmltools.py
@@ -97,8 +97,8 @@ def chunk(html: str):
     return list(document.generate_chunks())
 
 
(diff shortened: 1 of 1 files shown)


dgunning (approved):
Thanks for this change. Will integrate

PR #177, opened 2024-12-28
parse_company_facts sometimes errors out

rla3rd:
sometims "facts" in fjson is missing, and sometimes units_df is empty when processing facts from bulk data download


dgunning (approved):
Thanks, will integrate and test

PR #186, opened 2025-01-08
chore: add lru cache to filing html

jekozyra:
For our use case, we fetch both the html and the text. The text method calls the html method, so hopefully caching the call will lead to fewer SEC scraping requests.


jekozyra:
@dgunning I can't add reviewers for some reason but please take a look when you have a sec. Thanks!

dgunning (approved):
I guess this makes sense if you need both text and html in a quick call one after the other. Looks straightforward. Will approve and test

PR #189, opened 2025-01-10
Update attachments.py

rasmus393:
I think this structure is more logical at least before you could run into the issue that since it didnt check archive first it simply just ignored the flaged and then extracted into the directory unzipped which is unintuitive


dgunning (approved):
Thanks for this change. Seems ok. Approved

PR #191, opened 2025-01-13
chore: logging namespace to __name__ and defer logging initialization

### LINKED ISSUE #190: Logging namespace and top-level initialization
There are a few things in core.py that are problematic for applications that use edgartools as a library. 

-- Edit: 
This gets in the way of my usage, because at an unavoidable top-level it sets logging handlers, and uses "rich" as the logger rather than "edgartools". 

To quote https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library:
"use a logger with a unique and easily identifiable name, such as the __name__ for your library’s top-level package or module."
"Note It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers."
-- Edit
# Logger namespace
Core.py declares the logger as: 
https://github.com/dgunning/edgartools/blob/d4c47acc16d1c2ff54a0a512a6896d85f7ddff5d/edgar/core.py#L36

The result is that one can't set log levels explicitly on "edgartools", and edgartools is using rich's namespace for logging. 

Instead, this should be: 
`log = logging.getLogger(__name__)`

# Logging is initialized globally, when it should be deferred to the calling application
https://github.com/dgunning/edgartools/blob/d4c47acc16d1c2ff54a0a512a6896d85f7ddff5d/edgar/core.py#L29-L34

https://github.com/dgunning/edgartools/blob/d4c47acc16d1c2ff54a0a512a6896d85f7ddff5d/edgar/core.py#L44-L45

This setup should be deferred to the calling application. Could these be moved to a helper function invoked optionally... or removed entirely.

paultiq:
Fixes #190

- Change core.py logger to use `__name__` namespace, not "rich". 
- Move logger initialization and any additional log defaults, such as changing httpx log level, to `initialize_rich_logging`

This is consistent with recommendations from https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library:
- "use a logger with a unique and easily identifiable name, such as the `__name__` for your library’s top-level package or module."
- "Note It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers."

### DIFF
--- a/edgar/core.py
+++ b/edgar/core.py
@@ -25,24 +25,13 @@ from pandas.tseries.offsets import BDay
 from rich.logging import RichHandler
 from rich.prompt import Prompt
 
-# Rich logging
-logging.basicConfig(
-    level="INFO",
-    format="%(message)s",
-    datefmt="[%X]",
-    handlers=[RichHandler(rich_tracebacks=True)]
-)
-
-log = logging.getLogger("rich")
+log = logging.getLogger(__name__)
 
 # Pandas version
 pandas_version = tuple(map(int, pd.__version__.split('.')))
 
 # sys version
 python_version = tuple(map(int, sys.version.split()[0].split('.')))
-
-# Turn down 3rd party logging
-logging.getLogger("httpx").setLevel(logging.WARNING)
 
 __all__ = [
     'log',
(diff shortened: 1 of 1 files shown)


dgunning (approved):
Approved

PR #192, opened 2025-01-14
chore: Consolidate httpx.Client and httpx.AsyncClient creation to a single module

paultiq:
The current `httprequests` implementation creates httpx.Client and httpx.AsyncClient instances at multiple points. This scattered initialization makes it challenging, if not impossible, for downstream applications to customize httpx client behavior. 

This PR introduces a minimal change by consolidating all client initialization into a single location within a new `httpclient` module. This centralization simplifies the code and paves the way for downstream users to override or extend client functionality.

Examples of possible downstream changes include enabling client reuse, using [http2](https://www.python-httpx.org/http2/), implementing [caching](https://hishel.com/userguide/), handling custom responses, or supporting alternative rate-limiter implementations. 

I am proposing a minimal, non-disruptive change that doesn't change the current behavior. That said, client reuse has shown **significant** performance benefits for my use cases.

### DIFF
--- a/edgar/core.py
+++ b/edgar/core.py
@@ -452,29 +452,6 @@ def filter_by_ticker(data: pa.Table,
 def client_headers():
     return {'User-Agent': get_identity()}
 
-
-def http_client():
-    return httpx.Client(headers=client_headers(),
-                        timeout=edgar_mode.http_timeout,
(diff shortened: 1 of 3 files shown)


dgunning (approved):
Reviewed. Will integrate and test

PR #196, opened 2025-01-16
feat: Implement httpx client persistence / reuse

### LINKED ISSUE #195: Reusing HTTPX Clients
This follows from PR #192 , which consolidated the httpx client connection to a single edgartools.httpclient module. 

## Summary
edgartools would gain significant performance benefits from reusing HTTPX clients across requests, rather than creating a new client for each request. 

Reusing HTTPX client connections yields about a 20ms savings per request in retrieval in my, so far, limited tests. Results are shown below. 

I'll submit a PR shortly, but wanted to submit these initial results and ask: Are there additional test cases or considerations I should look at? 

## Current State
edgartools currently creates a new client *for each request*. This is problematic, because it results in separate HTTP connections for each and every request. 

This has a significant performance impact: testing (below) shows a significant speedup in retrieval. Pure retrieval tasks get ~20ms speedup by reusing a single HTTPX connection. 

## Solutions
There are several possible solutions here:
- Modify edgartools.httpclient (per #192) to serve / reuse a single HTTP client. I've implemented this and show results below.
- Require applications to explicitly create and maintain the httpclient, and change the API interfaces accordingly. This is fairly disruptive
- Change nothing: use cases that involve a small number of files retrieved at a time, or are otherwise CPU dominated, won't notice any impact.
- Add some hybrid option to enable/disable reuse on the fly. This would get complicated in threaded applications. 

To discuss in the PR: what behavior should be default? I'll propose a feature flag to enable/disable reuse. 

## More information

This is discussed in https://www.python-httpx.org/advanced/clients/

> When you make requests using the top-level API as documented in the [Quickstart](https://www.python-httpx.org/quickstart/) guide, HTTPX has to establish a new connection for every single request (connections are not reused). As the number of requests to a host increases, this quickly becomes inefficient.
> On the other hand, a Client instance uses [HTTP connection pooling](https://en.wikipedia.org/wiki/HTTP_persistent_connection). This means that when you make several requests to the same host, the Client will reuse the underlying TCP connection, instead of recreating one for every single request.

edgartools does use a client and does use context managers. The problem is the scope is per request, not across a set of requests: 
> The recommended way to use a Client is as a context manager. This will ensure that connections are properly cleaned up when leaving the with block

## Implementation Notes
For synchronous code, this is straightforward: create a single HTTPX Client, and serve it, until it is closed. 

For async code, this requires a separate AsyncClient per thread, since you can't reuse the client across event loops.

The side effect is: The client would remain "open" until closed by the caller. 

#### Current: get_filings
|Test Case|Function|Iteration|Seconds|
|-|-|-|-|
|REUSE=False|get_filings|0|0.38|
|REUSE=False|get_filings|1|0.29|
|REUSE=False|get_filings|2|0.27|
|REUSE=False|get_filings|3|0.29|
|REUSE=False|get_filings|4|0.28|

A new client is create for each connection. 

#### Reusing Client: get_filings
|Test Case|Function|Iteration|Seconds|
|-|-|-|-|
|REUSE=True|get_filings|0|0.30|
|REUSE=True|get_filings|1|0.12|
|REUSE=True|get_filings|2|0.11|
|REUSE=True|get_filings|3|0.11|
|REUSE=True|get_filings|4|0.11|

The first request (Iteration 0) results in a new Client. Subsequent requests are faster.

#### Current: _download_filing_text first 20 
|Test Case|Function|Iteration|Seconds|
|-|-|-|-|
|REUSE=True|test_retrieve_20|0|5.56|
|REUSE=True|test_retrieve_20|1|5.62|
|REUSE=True|test_retrieve_20|2|5.99|

#### Reusing Client: _download_filing_text first 20 
|Test Case|Function|Iteration|Seconds|
|-|-|-|-|
|REUSE=False|test_retrieve_20|0|8.30|
|REUSE=False|test_retrieve_20|1|8.05|
|REUSE=False|test_retrieve_20|2|8.01|

#### Adhoc Pytest comparisons:
|Test|Current (sec)|Reusing Client (sec)|
|-|-:|-:|
|tests/test_company.py|12.10|7.5|
|tests/test_xbrl_10Q.py|6.63|6.09|
|tests/test_httprequests|2.29|1.84|
|tests/test_filing.py|51.82|41.05|

##### test_retrieve_20 test case

I have this in a separate testing notebook right now, which I may later add to the PR, but the rough code for this is: 
```py

filings = get_filings()
def retrieve_filings(filings, limit: int):
    for i, f in enumerate(filings):
        f._download_filing_text()
        if i>limit:
            break

retrieve_filings(filings, 20)
```

paultiq:
Implements a feature flag `edgartools.httpclient.PERSISTENT_CLIENT`. 

When disabled, the flag should have no functional impact on edgartools: a new HTTPX client would be created for each and every request. No change from prior behavior.

When enabled, the client would be reused across all requests. For synchronous requests, a single client is reused. For async requests, a client is attached and used for each asyncio loop. 

If enabled, an application would (if desired) need to explicitly close the client.

A parameter dictionary is added `httpclient.DEFAULT_PARAMS`, so that applications can modify the httpclient parameters used on client initialization. For example, could set  `httpclient.DEFAULT_PARAMS["http2"] = True` to enable HTTP2 mode in httpx.

Note: One concern is whether garbage collection of the async clients is desirable or whether to require explicit closure. As implemented, the async clients are maintained in a WeakKeyDictionary, so will be gc'd eagerly. 

Fixes #195

### DIFF
--- a/edgar/core.py
+++ b/edgar/core.py
@@ -25,6 +25,8 @@ from pandas.tseries.offsets import BDay
 from rich.logging import RichHandler
 from rich.prompt import Prompt
 
+from edgar.httpclient import close_clients
+
 log = logging.getLogger(__name__)
 
 # Pandas version
@@ -171,6 +173,8 @@ def set_identity(user_identity: str):
     """
     os.environ[edgar_identity] = user_identity
     log.info(f"Identity of the Edgar REST client set to [{user_identity}]")
+
+    close_clients() # close any httpx clients, to reset the identity. 
 
 
 identity_prompt = """
--- a/edgar/core.py
+++ b/edgar/core.py
@@ -25,8 +25,6 @@ from pandas.tseries.offsets import BDay
 from rich.logging import RichHandler
 from rich.prompt import Prompt
 
-from edgar.httpclient import close_clients
-
 log = logging.getLogger(__name__)
 
 # Pandas version
@@ -174,6 +172,7 @@ def set_identity(user_identity: str):
     os.environ[edgar_identity] = user_identity
     log.info(f"Identity of the Edgar REST client set to [{user_identity}]")
 
+    from edgar.httpclient import close_clients
     close_clients() # close any httpx clients, to reset the identity. 
 
 
--- a/edgar/httpclient.py
+++ b/edgar/httpclient.py
@@ -2,15 +2,14 @@ import logging
 import threading
 import asyncio
 from contextlib import asynccontextmanager, contextmanager
-
+from weakref import WeakKeyDictionary
 import httpx
 
 from edgar.core import client_headers, edgar_mode
 
-
 log = logging.getLogger(__name__)
 
-REUSE_CLIENT = True
+PERSISTENT_CLIENT = True # When enabled, httpclient reuses httpx clients rather than creating a single request per client
 
 DEFAULT_PARAMS = {
     "timeout": edgar_mode.http_timeout,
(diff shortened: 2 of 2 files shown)


dgunning (commented):
I understand the changes. When were you proposing to turn the feature on? Is putting it behind the feature flag just for discussion?

paultiq:
The feature flag was primarily for discussion and testing, since it's easy to toggle it on and off and do comparisons. 

While I believe a persistent client has enough benefits where it should be the default, I wanted to throw this draft out there to see if you / anyone had concerns about this change. 

A few things that are a work-in-progress:
- Locking to avoid creating multiple connections if the first request is parallel().
- The user's identity is set in the headers of the Client. Therefore, set_identity should also "close" the http clients (if already open), so they'll be recreated with the new identity. 
- Test and clean up the AsyncClient, and use either a ContextVar or attach to the asyncio loop rather than the thread.

** Separately, I'm looking at the throttle_requests implementation. I think there's another easy gain to be made from switching to a different throttling behind an httpx transport.

paultiq:
I think this is ready for review.

In my opinion, the benefits of a persistent client are significant enough to merit making it the default, as committed: `PERSISTENT_CLIENT=True`. This could be moved behind an environment variable... or just made default. 

Related to these changes, but not in the PR, are two ideas I have implemented on another branch: 
- Using an httpx transport to replace the throttle_requests implementation
- Using an httpx caching library [hishel](https://pypi.org/project/hishel/) , as an alternative to the local storage. Some additional rules would be needed to identify which URLs are cacheable vs not.

paultiq:
Note: the 'more correct' approach for async is to change every path that uses async to create and close a client and pass it to each task, rather than for each request. 

I opted to do this implicitly, to avoid too substantial a change... but it's worth considering making the callers responsible for client creation.

paultiq:
Sorry... now that I think about this, I have an alternative solution I'm cooking.

The correct approach is to just fix the async code, and it's not as bad as I was thinking. I'll show both versions shortly.

PR #197, opened 2025-01-19
feat; Persistent HTTP Clients

paultiq:
When disabled, the flag should have no functional impact on edgartools: a new HTTPX client would be created for each and every request. No change from prior behavior.

When enabled, the client would be reused across all requests. For synchronous requests, a single client is reused. For async requests, a client is attached and used for each asyncio loop.

If enabled, an application would (if desired) need to explicitly close the client.

A parameter dictionary is added httpclient.DEFAULT_PARAMS, so that applications can modify the httpclient parameters used on client initialization. For example, could set httpclient.DEFAULT_PARAMS["http2"] = True to enable HTTP2 mode in httpx.

Note: One concern is whether garbage collection of the async clients is desirable or whether to require explicit closure. As implemented, the async clients are maintained in a WeakKeyDictionary, so will be gc'd eagerly.

Note 2: This differs from #196 in the Async code path. In #196, I avoided touching the code outside httpclient / httprequests. This ended up with an unnatural solution: where AsyncClients may or may not be closed. In this, the async code paths are modified to take an optional client... and the top level asyncio.gather's are modified to create a single AsyncClient for the entire set. This is cleaner and more explicit.

### DIFF
--- a/edgar/attachments.py
+++ b/edgar/attachments.py
@@ -26,6 +26,7 @@ from rich.text import Text
 from edgar.richtools import repr_rich, print_xml, print_rich
 from edgar.core import sec_dot_gov, display_size, binary_extensions, text_extensions, has_html_content
 from edgar.httprequests import get_with_retry, download_file, download_file_async
+from edgar.httpclient import async_http_client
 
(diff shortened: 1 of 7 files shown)


dgunning (approved):
Approved. Will merge and do thorough testing

PR #206, opened 2025-01-23
Use `exists_ok` over `exists()` for edgar_data_dir

hspak:
We are hitting an `FileExistsError` exception despite the `exists()` check. Opt to use the [`exists_ok`](https://docs.python.org/3/library/os.html#os.makedirs) parameter for `makedirs` to prevent the exception.

I suppose we could keep the `exists()` check as well, but I removed because it felt redundant. Happy to change.

Example case:
Screenshot 2025-01-23 at 2 45 30 PM


dgunning (approved):
Approved

PR #214, opened 2025-02-02
feat: Request caching using Hishel / HTTPX

paultiq:
As a follow-up to https://github.com/dgunning/edgartools/discussions/179#discussioncomment-12033238, sharing this draft PR for discussion.

#### Caching
- Introduces `edgar.httpclient_cache` module
- Cache is activated via `edgar.httpclient_cache.install_cached_client(cache_directory, controller_args)`
- To force caching of everything, pass `controller_args = {"force_cache": True}`
- To use an [RFC 9111](https://www.rfc-editor.org/rfc/rfc9111.html#name-calculating-heuristic-fresh) heuristic cache (caches at min(10% of file age, 1 week), pass `controller_args = {"allow_heuristics": True, "allow_stale": True, "always_revalidate": False}`

Note: Alternative caching policies could be implemented, but defining a cache controller with custom rules... such as certain URLs for X time. That said, I think the heuristic rules are a good start - data that's less than a day old would only be cached for an hour. 

#### Rate Limiting
Rate limiting must be done at the request level, not at the application layer, so that only cache misses are throttled. 

- Added an override to edgar.httprequest to disable the @throttle_requests. This is activated when `install_cached_client` is called.
- Decorate the HTTPX transport with a rate limiter, using the same limiter across all clients (Async or Sync)

#### New dependencies
Optional - only needed if you use the httpclient_cache module. 
- https://pypi.org/project/limiter/
- https://pypi.org/project/hishel/

#### Testing notes
Since this is disabled by default, the test cases currently won't exercise this feature.

### DIFF
--- a/edgar/httpclient.py
+++ b/edgar/httpclient.py
@@ -16,13 +16,16 @@ DEFAULT_PARAMS = {
     "default_encoding": "utf-8",
 }
 
+client_factory_class = httpx.Client
+asyncclient_factory_class = httpx.AsyncClient
+
 def _client_factory(**kwargs)-> httpx.Client:
     params = DEFAULT_PARAMS.copy()
     params["headers"] = client_headers()
     
     params.update(**kwargs)
     
-    return httpx.Client(**params)
+    return client_factory_class(**params)
 
 def _http_client_manager():
     """When PERSISTENT_CLIENT, creates and reuses a single client. Otherwise, creates a new client per invocation."""
@@ -76,7 +79,7 @@ async def async_http_client(client: Optional[httpx.AsyncClient] = None, **kwargs
     params["headers"] = client_headers()
     
     params.update(**kwargs)
-    async with httpx.AsyncClient(**params) as client:
+    async with asyncclient_factory_class(**params) as client:
         yield client
 
 def close_clients():
--- a/edgar/httpclient_cache.py
+++ b/edgar/httpclient_cache.py
@@ -41,7 +41,7 @@ import httpx
 
 logger = logging.getLogger(__name__)
 
-LIMITER = Limiter(rate=10, units=1)
+LIMITER = Limiter(rate=10, capacity=10, units=1)
 
 def custom_key_generator(request: httpcore.Request, body: bytes | None) -> str:
     """ Generates a stable, readable key for a given request.
@@ -59,7 +59,9 @@ def custom_key_generator(request: httpcore.Request, body: bytes | None) -> str:
     url = request.url.target.decode()
 
     url_p = url.replace("/", "__")
-    return f"{host}_{url_p}"
+
+    key = f"{host}_{url_p}"
+    return key
 
 def _get_cache_controller(**kwargs):
     controller = hishel.Controller(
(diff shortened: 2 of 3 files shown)


dgunning:
Starting to look at this now. 

How granular is the configuration for urls? The data behind **https://www.sec.gov/Archives/edgar/full-index/2024/QTR4/form.idx** gets updates every night at 10:30 even though the url stays the same

paultiq:
As granular as you want. There's no per-URL logic implemented, by default is just uses request & response headers. You'd implement a custom CacheController that interprets the request (& last response) however you want. 

You could also override the heuristic behavior. The default heuristic is 10% of file age (or 1 week). For Edgar, perhaps you'd say:  **Don't cache anything in /Archives/edgar/full-index that has an age of less than 24 hours.**

dgunning:
Everything under /Archives/edgar/full-index/2024/QTR4 and before should be cached
Everything under /Archives/edgar/full-index/2025/QTR1 should **not** be cached

Then on April 2025
Everything under /Archives/edgar/full-index/2025/QTR1 and before should be cached
Everything under /Archives/edgar/full-index/2025/QTR2 should **not** be cached

paultiq:
Understood... I was just trying to think of a rule that didn't require URL parsing, or required minimal logic. 

That's fine tho, I'll stub out a framework for this, so additional rules can easily be added.

dgunning:
What are the mechanics of the current changes? When we turn on the cache with `edgar.httpclient_cache.install_cached_client(cache_directory, controller_args)` does it cache every request? Is it off by default?

paultiq:
As is:

Depends on the controller args. If you pass no args, I believe nothing ends up caching because Edgar doesn't pass any caching response headers.

There's examples in httpclient_cache, but I gave examples for either force cache (caches everything), or heuristic cache (caches up to min(1 week, 10% of age)).

paultiq:
Yes: as it stands, it's a conservative PR that leaves it to the user to decide on cache policy.

I guess the question is whether to wait for a cache controller that incorporates any Edgar specific rules, or ship this as an advanced feature.

I'm fine with both options... whatever you'd prefer

dgunning:
I don't see any changes to pyproject.toml. Are you planning on testing this for a period before general release?

paultiq:
This is one of the open questions. Consolidating: 
1. How to handle the additional [but optional] dependencies for caching? Listed in initial comment above. 
2. Should we hold the PR until we have an edgar cache policy/controller, or ship this PR and implement that separate? 
3. Testing - How to modify test cases to, at least, make sure nothing breaks when caching is enabled?

dgunning:
Have you been testing using the limiter package instead of the one that is in the edgartools code? How is that testing going?

paultiq:
Yes, I've been testing with caching enable using the limiter package. So far, so good... I hope to have some time later this week to stress test it. 

If you're comfortable with the PR, but without a "custom controller" (for Edgar rules), I'll do another review and figure out how to run the test cases with caching enabled.

dgunning:
That would be OK

paultiq:
Finally was able to set aside some time. The rate limiting seems to work well, I did a few tests with much larger rate limits to ensure I hit the 429/too many requests, but overall this is fairly straightforward.

dgunning:
Hey, I was tied up in a major rewrite of XBRLs and Financials. Will look at this later tonight

dgunning (approved):
Approved

PR #253, opened 2025-03-25
Fix bug where press release url's are not properly printed

zero-stroke:
This used to work fine, but updates in late January broke this feature. The last commit where this worked fine was 345aa048a30ebbe76bdf4507da8250fce7f2d84c , "Add uu_decode to convert SGML attachments to bytes".

Uses f string to properly set the url

This is the script I was using to test if the url's were properly being printed: 

```py
import edgar
from edgar import Company, set_identity, company_reports

set_identity("my_email@gmail.com")

company = Company('A')
start_date_str = '2024-01-01'

all_filings = company.get_filings(form=['8-K'], date=f"{start_date_str}:")
all_filings = list(all_filings)

for filing in all_filings[:4]:
    all_prs: list[edgar.company_reports.PressRelease] = filing.obj().press_releases

    if not all_prs:
        continue

    print(all_prs)

    for pr in all_prs:
        print(pr.url())
        print(pr.attachment.path)
        print('\n')
        # print(pr)
```

If you run it with the current version you'll see the url's aren't printed properly. If you run it with this PR or with versions older than Jan 25th you'll see they are printed properly.

### DIFF
--- a/edgar/sgml/sgml_common.py
+++ b/edgar/sgml/sgml_common.py
@@ -188,7 +188,6 @@ def parse_submission_text(content: str) -> Tuple[FilingHeader, DefaultDict[str, 
     return header, documents
 
 
-
 class FilingSGML:
     """
     Main class that parses and provides access to both the header and documents
@@ -230,7 +229,6 @@ class FilingSGML:
     def effective_date(self):
         return self.header.filing_metadata.get('EFFECTIVE DATE')
 
-
     def html(self):
         html_document = self.attachments.primary_html_document
         if html_document and not html_document.is_binary() and not html_document.empty:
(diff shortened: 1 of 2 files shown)


dgunning:
We switched from using the filing home page to using the SGML text documents in January since most attachments on some filings are in the SGML but not listed on the homepage. The downside of using SGML attachments is that some attachments will have urls that will return 404 not found errors and so the url was changed for all attachments to indicate they came from SGML. That is the root cause if the bug that this PR fixes.

The challenge with adding back the urls to all attachments will be the 401 issue for some attachments. I wonder if we add it will some code run into 404s? Is this just an issue for Press Releases? I'll look into the code then check what the impact will be

zero-stroke:
I've only had this come up for press release urls, not sure what other url's could be affected. There was little actual code changes needed for this to be fixed.

dgunning (commented):
Approved. Might make a minor modification to use existing utility functions.

dgunning:
@zero-stroke thanks for the pull request. Because of the old filings I had to apply a patch locally and effectively apply your changes. I can't figure out how to get your pull request to pass the test and I guess I could ask you for changes. However what I will do is effectively apply your changes locally and test, and close your pull request.
Hope this works

zero-stroke:
@dgunning Aw I was looking forward to being on the contributors list D:

PR #255, opened 2025-03-30
1. Error: "can't multiply sequence by non-int of type 'float'"

nihaoo:
2. Root cause: Sometimes, row.Shares or row.UnderlyingShares value is polluted by footnote reference, for example, '923 F1', in old version filing form. Need to remove the footnote reference, 'F1' and return pure share value, 923 in this example.
3. Test case: a) accession_no: 0000899243-15-008442
              b) accession_no: 0000950170-24-114345


dgunning (approved):
Approved

PR #262, opened 2025-04-09
Bugfix: Reading Index Files

sean-enderby-wbs:
Addresses two issues with the parsing of SEC index files:

1. When reading form indexes, parts of form names which contain spaces were included in the company name. For example, where the form name is "1-A POS" the string "POS" was prepended to all company names.
2. When reading company indexes only the first 12 characters of each line of the index were used as the company name. The remaining text was incorporated into the form name, creating many erroneous form names in the output.

_tests/test_read_filing_indexes.py_ also updated to ensure that form names are read correctly from both types of index file.


dgunning (approved):
Approved. Will test. Thanks

PR #300, opened 2025-05-29
fix: Documentation Updates

ChuckBasin:
1) Remove def from download_edgar_data method call
2) Update latest_10k to get_filings in xrbl.md to match latest methods


dgunning (approved):
Approved. Thanks a lot

Issue #297, opened 2025-05-26
Allow disabling SSL verification for HTTPX requests


tybeller:
Hello,

Was wondering if there would be any support in allowing users to disable SSL verification on HTTPX requests, mainly with regards to pulling filings.

Can save a bit of a headache in corporate networks.

Having just taken a peek at the code, seems like it should be as simple as just passing the arg in get_with_retry and get_with_retry_async and following that upstream to where its exposed in functions to allow the user to set it wherever it gets called.

Would be happy to take a crack at it, Thanks!

dgunning:
Hello,

I don't know if there is anyplace upstream to set verify=False.

What you could do instead is use an environment variable e.g.  `EDGARTOOLS_VERIFY_SSL` and add  the following line to the `DEFAULT_PARAMS` dict in httpclient.py

`"verify": os.environ.get("EDGAR_VERIFY_SSL", "true").lower() != "false"`,

I think that would be the extent of the changes. We would then need to document and test

dgunning:
Closing this issue. Raise it as a PR when ready. Thanks

PR #303, opened 2025-05-31
feat: Add support for setting SSL verification in http requests

tybeller:
Added the http request ssl verification mentioned in #297, as well as a unit test to verify.

Wasn't sure where this change fits in the docs, read through and couldn't find anywhere that really made sense, but its also such a small add it doesn't feel right to make a whole new page.

Additionally, I see in `edgar/storage.py`, the `EDGAR_USE_LOCAL_DATA` env variable can be set as 1 or 0 or True or False etc. (but does not handle the case where the env variable is missing which is necessary for this). Based on the [suggestion from the issue](https://github.com/dgunning/edgartools/issues/297#issuecomment-2910685578) , `EDGAR_SSL_VERIFY` Only resolves to false when the environment variable is the word false.

Thanks!

### DIFF
--- a/edgar/httpclient.py
+++ b/edgar/httpclient.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import threading
 from contextlib import asynccontextmanager, contextmanager, nullcontext
 import httpx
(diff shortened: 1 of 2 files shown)


ChuckBasin:
Thoughts on changing:
```
    return bool(os.environ.get("EDGAR_VERIFY_SSL") == "True")
```

for [EDGAR_VERIFY_SSL](https://github.com/dgunning/edgartools/pull/303/files#diff-3356a887bae7983f0dc1e3f72d1fd2a66a2a986a6f01c09a55370599835cfefdR15)

Pythonic way seems to be case "True/False" when variable assignment to boolean. This may make sense to enforce on environment variables at the OS level to remove confusion. Ie. an os env variable is either True or False and not allowing to .lower() to allow true/false.

dgunning (approved):
Reviewed the changes. I think they are appropriate for the feature with minimal side effects. Will approve then do some additional testing

PR #318, opened 2025-06-10
Silence uu warnings

### LINKED ISSUE #317: Silence uu warnings about Trailing Garbage
When processing multiple SGML files in a row, the console gets filled with `Warning: Trailing Garbage` messages. This seems to be originating from the `uu` module, which is only used in one place: 

https://github.com/dgunning/edgartools/blob/45f0ac04604528c6d15bfa4ead30851336a675e7/edgar/sgml/sgml_parser.py#L51-L52

Specifically, the `uu.decode()` function takes a `quiet` parameter with a default value of `False`:

https://github.com/dgunning/edgartools/blob/45f0ac04604528c6d15bfa4ead30851336a675e7/edgar/vendored/uu.py#L100

The only place it is used in the function is in this snippet:

https://github.com/dgunning/edgartools/blob/45f0ac04604528c6d15bfa4ead30851336a675e7/edgar/vendored/uu.py#L157-L166

There is no benefit to displaying the error since the workaround handles it. It looks like there was already an intent to silence these warnings:

https://github.com/dgunning/edgartools/blob/45f0ac04604528c6d15bfa4ead30851336a675e7/edgar/sgml/sgml_parser.py#L45

I see two ways of resolving this issue:

1. Change the call to `uu.decode()` in `sgml_parser.py` to use quiet mode. I think this fix makes the most sense.
2. Changed the line that writes the warning in `uu.py` to use the logging module to log the warning to the `DEBUG` level. This would allow the message to get through when debugging, but it implies modifying a vendored module embedded as-is in edgartools.

I'm happy to write a fix using the preferred approach.

vgreg:
Call `uu.decode()` with `quiet=True` to silence the unnecessary `Warning: Trailing Garbage` messages. Fixes #317

### DIFF
--- a/edgar/sgml/sgml_parser.py
+++ b/edgar/sgml/sgml_parser.py
@@ -49,7 +49,7 @@ class SGMLDocument:
                 output_stream = BytesIO()
 
                 # Decode the UU content
-                uu.decode(input_stream, output_stream)
+                uu.decode(input_stream, output_stream, quiet=True)
 
                 # Get the decoded bytes
                 return output_stream.getvalue()


dgunning (approved):
Thanks for this change. Approved

PR #340, opened 2025-07-07
Add markdown conversion support for HTML attachments and add support for optional page breaks

MauritsBrinkman:
This PR adds markdown conversion support for HTML attachments, optional page breaks in markdown output and fixes parsing issues with nested HTML structure. Practical use case for this is when needing documents in machine-readable format, and being able to correctly match by page.

## New Features

- **Markdown conversion**: Added `markdown()` method to `Attachment` and `Attachments` classes
- **Page break support**: Added optional `include_page_breaks` parameter with delimiter format `{page_number}------------------------------------------------` in markdown output

## Bug Fixes

- **Improved HTML parsing**: Now properly handles nested `<tbody>`, `<thead>`, and `<tfoot>` elements. With previous code, some content in the markdown output was completely skipped, as only direct children were considered

### DIFF
--- a/ai_docs/attachments_api.md
+++ b/ai_docs/attachments_api.md
@@ -70,8 +70,33 @@ Each `Attachment` provides rich metadata and content extraction methods:
 - `attachment.description`: description from SEC filing
 - `attachment.document_type`: type (e.g., 'EX-99.1', 'HTML', etc.)
 - `attachment.text()`: plain text content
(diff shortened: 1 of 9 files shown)


MauritsBrinkman:
Hey @dgunning ! Hope you are doing well! Submitting this for review :)

MauritsBrinkman:
Yes, I think that's simply because identity is not set? I'll have a look tomorrow and see!

PR #346, opened 2025-07-10
Save on disk

danielerosa:
Writing directly to disk while downloading instead of keeping it in memory.
Note the the Path().rename() might break across different file systems.
Standard formatting with isort and black. You could get rid of them if you don't care.

### DIFF
--- a/edgar/httprequests.py
+++ b/edgar/httprequests.py
@@ -10,29 +10,43 @@ from functools import wraps
 from io import BytesIO
 from pathlib import Path
 from threading import Lock
-from typing import Union, Optional
+from typing import Optional, Union
 
-from httpx import RequestError, Response, AsyncClient
 import orjson as json
+from httpx import AsyncClient, RequestError, Response
 from stamina import retry
 from tqdm import tqdm
 
-from edgar.core import text_extensions, get_edgar_data_directory
-from edgar.httpclient import http_client, async_http_client
+from edgar.core import get_edgar_data_directory, text_extensions
+from edgar.httpclient import async_http_client, http_client
 
 """
 This module provides functions to handle HTTP requests with retry logic, throttling, and identity management.
 """
-__all__ = ["get_with_retry", "get_with_retry_async", "stream_with_retry", "post_with_retry", "post_with_retry_async",
-           "download_file", "download_file_async", "download_json", "download_json_async", "stream_file",
-           "download_text", "download_text_between_tags", "download_bulk_data", "download_datafile",
-           "throttle_requests"]
+__all__ = [
+    "get_with_retry",
+    "get_with_retry_async",
+    "stream_with_retry",
+    "post_with_retry",
+    "post_with_retry_async",
+    "download_file",
+    "download_file_async",
+    "download_json",
+    "download_json_async",
+    "stream_file",
+    "download_text",
+    "download_text_between_tags",
+    "download_bulk_data",
+    "download_datafile",
+    "throttle_requests",
+]
 
 attempts = 6
 retry_timeout = 40
 wait_initial = 0.1
 max_requests_per_second = 8
 throttle_disabled = False
+
 
 class TooManyRequestsError(Exception):
     def __init__(self, url, message="Too Many Requests"):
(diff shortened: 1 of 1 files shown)


dgunning:
Let's start a conversation for this PR. I see it is about saving to disk  if the file size is greater that 2GB. Do you have examples of downloads greater than 2GB? What happens during those downloads?

danielerosa:
hi @dgunning , btw, thank you for this work. Very good quality.

So I have been targeting 1 day, a few forms
            "4",
            "4/A",
            "8-K",
            "8-K/A",
            "SC 13G",
            "SC 13G/A",
            "SCHEDULE 13G/A"
, and 2 exchanges, nyse and nasdaq

image

then I call you download_filings(filings=ifilings) and I get files relatively big sometimes:
image

does it check out with you?

I figure that instead of using memory I could use the file system directly.

To your question, what happens during those downloads. You just use memory and it's not for a great cause given that you don't need speed as you are reading from network, slower than filesystem.

dgunning:
I did a code review with **Claude Code** of the current `stream_file` function in **edgartools**. 

In summary we should change to using disk, but always use disk for all fie sizes. 
Also from my earlier comment we should use `tempfile` instead of `/tmp`

## Claude Code Review

**Question:**  _Review edgar.httprequests.stream_file. Give me your thoughts on downloading really large files like 2GB which can happen with bulk edgar data. Would downloading to a temp directory help?_

 ### Current Problems

  1. **Memory accumulation**: The function loads the entire file into memory (content = b'' then content += chunk) before writing to disk, requiring 2GB+ RAM for large files.
  2. **No partial recovery**: If the download fails partway through, all progress is lost since nothing is written until the end.
  3. **Memory pressure**: Large downloads could cause system memory issues or OOM errors.

  ### Temp Directory Benefits

  Yes, downloading to a temp directory would significantly help:

  1. **Stream-to-disk**: Write chunks directly to a temporary file as they arrive, keeping memory usage constant (~4MB for the chunk buffer).
  2. **Atomic operation**s: Download to temp, then move to final location when complete, avoiding partial files in the target directory.
  3. **Resume capability**: Temp files could potentially support resume functionality.
  4. **Error safety**: Failed downloads are automatically cleaned up from temp without affecting the target directory.

  The current approach of `content += chunk` followed by `save_or_return_content(content, path) `at line 563 is problematic for bulk data downloads. A streaming approach writing directly to disk would be
  much more robust and memory-efficient.

  For a user's PR using the file system, this would be a valuable improvement to support the large SEC bulk data files without memory constraints.
  
  **Question:** _Does it make sense to use disk over a certain size OR just use disk when streaming_
  
  ### Always Stream to Disk

  The current mixed approach (memory for small files, stream for large) adds unnecessary complexity. When you're already streaming (aiter_bytes), you should write directly to disk regardless of size.

  ### Benefits of Always-Disk Approach

  1. **Consistent behavior**: No arbitrary size thresholds to maintain
  2. **Simpler code**: Single code path eliminates conditional logic
  3. **Memory efficiency**: 4MB chunk buffer vs potentially gigabytes in memory
  4. **Future-proof**: Works for any file size without modification

  ### Current Issues

  The stream_file function is already streaming chunks but then concatenating them in memory:
  ```python
  content = b''
  async for chunk in response.aiter_bytes(chunk_size=CHUNK_SIZE):
      content += chunk  # This defeats the purpose of streaming
```

  ### Better Approach

  Since you're streaming anyway, write chunks directly to a temp file, then move to final location. This gives you atomic operations and constant memory usage regardless of file size.

  The overhead of disk I/O for small files is negligible compared to the memory safety and code simplicity benefits. Modern SSDs make this even less of a concern.

  For the PR review, I'd recommend suggesting they always stream to disk when using the streaming function - it's the primary reason to use streaming over regular download.

dgunning:
I use Claude Code and I trust it with important bits of code like the `stream_file` function, and in theory I could just ask it to fix the function, but I'm trying to balance with supporting community involvement. So if you want to proceed with this PR make the suggested changes. I appreciate that you brought this up

danielerosa:
@dgunning I could refine this PR but I will be on travel for work for a few days and this is technically "not work". So I could take a look when I am back.
Also, what's you preference regarding formatting? I am using default isort and black from nvim setup.

dgunning:
> @dgunning I could refine this PR but I will be on travel for work for a few days and this is technically "not work". So I could take a look when I am back. Also, what's you preference regarding formatting? I am using default isort and black from nvim setup.

I could also just accept the PR and make the quick edits. You would be listed as a contributor and I would be grateful

danielerosa:
good with me @dgunning either way

danielerosa:
hi @dgunning, sorry I missed the :point_up: . I was on the road. It looks like you did it all. Thank you!

PR #347, opened 2025-07-11
Check and optimize the item distinction function

zhangchmeng:
Subsequent tests and checks will be conducted to continuously fix the item distinction function

### DIFF
--- a/edgar/files/html_documents.py
+++ b/edgar/files/html_documents.py
@@ -516,7 +516,10 @@ class HtmlDocument:
                     if current_chunk:
                         if any(block.text.strip() for block in current_chunk):  # Avoid emitting empty chunks
                             yield current_chunk
-                        current_chunk = []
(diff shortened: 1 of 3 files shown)


zhangchmeng:
> 

A considerable number of reports in 10-k 10-q do not always use 'item' as the beginning, but we can distinguish and capture different blocks from the links in the table of contents

zhangchmeng:
> What is the overall purpose of this pull request? What was the issue and how does it fix it?

A considerable number of reports in 10-k 10-q do not always use 'item' as the beginning, but we can distinguish and capture different blocks from the links in the table of contents

zhangchmeng:
> What is the overall purpose of this pull request? What was the issue and how does it fix it?

current code fix some report, Some reports directly use the 'a' tag as the top-level element instead of other block blocks

PR #377, opened 2025-08-08
Revert back to a single global limiter

paultiq:
This reverts back to original intent: a single global Limiter. 

The only issue is: pyrate_limiter will emit a warning message about "async call made without an async bucket". 

This is ignorable and the log message is now a .debug in pyrate_limiter master: 

https://github.com/vutran1710/PyrateLimiter/blob/a8fc3d4720493199b27b39f49d2629443c29db23/pyrate_limiter/limiter.py#L363


dgunning (commented):
I had also made this change https://github.com/dgunning/edgartools/blob/main/edgar/httpclient_ratelimiter.py#L20

Can you check and revert if required?

dgunning (approved):
Approved

Issue #378, opened 2025-08-09
Cache Rules


paultiq:
Does it make sense to cache bulk files, where edgartools already has an explicit download check? 

Examples: 
https://www.sec.gov/Archives/edgar/Feed/2024/QTR1/20240118.nc.tar.gz
https://www.sec.gov/Archives/edgar/daily-index/bulkdata/submissions.zip
https://www.sec.gov/Archives/edgar/daily-index/xbrl/companyfacts.zip

dgunning:
The answer to this is to probably let the user decide. I can see the value of caching the daily feed files - in Local Storage you can download only 10-Ks, then in another call you add 8-Ks, and the same file is downloaded and those filings saved. If we had caching then the second download is saved. But I don't know if users will expect for those large files to be cached, and effectively you will need double the storage. By default the setting should be no, but users can change

paultiq:
Ok, as is, they're caching twice. 

I'll come up with something to disable these by default.

PR #382, opened 2025-08-09
Cache Performance - Large Files

paultiq:
While reviewing #378, I noticed that bulk files were much slower with the cache enabled due to serialization:

This PR contains three performance optimizations:
- Use a faster [serializer ](https://hishel.com/advanced/serializers/) than the default JSON serializer
- Fixes revalidation - after the cache time is up, we should revalidate the data not re-download again. This needed the `_make_request_conditional` option. 
- Enable [HTTP2 ](https://www.python-httpx.org/http2/) if H2 is installed: I haven't yet seen a significant difference in my testing so far, but http2 may help complex apps with many concurrent operations.

I did some performance testing, and this cuts the cache hits for large files from 14 seconds to 1-2 seconds. 

More broadly, a set of performance tests will be helpful. Not sure yet how though.


dgunning (approved):
Approved

PR #392, opened 2025-08-12
pre-commits

paultiq:
@dgunning This is just a suggestion - I like pre-commits, and especially ruff. These help a bit as little formatting differences have crept in. Ruff shows a lot of format and lint warnings when I run against all files, but it'll only complain about staged files by default, so it'll be more of an incremental process. 

Setup - run this once in your repo directory locally - 
```py
pip install pre-commit
pre-commit install
``` 

Usage: 
pre-commit will run automatically on any git commit. You can override it with git commit --no-verify. 

You can run manually with `pre-commit run`, which runs against any staged files. 

`pre-commit run --all-files` checks all files. I just ran this, and it touched about 180 files, so up to you whether you want to do one big "format" commit or not.


dgunning (approved):
Approved

Issue #413, opened 2025-08-28
Feature Request: Create Filing object from full text submission


parijatgarg:
## Description

It is possible to create a Filing object by pointing to a file on disk or a URL. However, we would like to be able to create a Filing object directly from a FilingSGML object or a full text submission string.

## Why it would be beneficial

For various data pipelines, it is sometimes not possible to have the data on disk. It is obtained from a data store and we want to directly create the Filing object. Internally, it is already possible to create a Filing object using a FilingSGML object (but by providing a from_source url/file).

## Implementation ideas

It is straightforward to add a from_text method that takes a full_text_submission string. FilingSGML already supports this. So it is only a matter of replicating the Filing.from_source method to Filing.from_text and changing the call that is used to create the FilingSGML internally.

The tools are so powerful. Being able to integrate it into my data pipeline would be super useful.

EricPiazza:
Agreed, I would like this feature.

This package relies on HTTP (which isn't always desirable) and local storage (which isn't always available, such as in cloud environments).  I find the parsing features to be very powerful and useful but they are highly coupled with the retrieval features.  I would like to see more support for situations where the method of retrieving content is more flexible/agnostic.

Ideally, all object types should be able to be hydrated by string content.

dgunning:
Available in [Release 4.11.0](https://github.com/dgunning/edgartools/releases/tag/v4.11.0)

PR #414, opened 2025-08-28
Create Filing from full text submission - Issue #413

parijatgarg:
Adds a from_text method to Filing class. This lets one create a Filing object directly from a full text submission string instead of a url or file path. This was a feature request I raised as Issue #413 and I have taken the liberty of providing an implementation.

### DIFF
--- a/edgar/_filings.py
+++ b/edgar/_filings.py
@@ -1300,6 +1300,26 @@ class Filing:
         filing._sgml = filing_sgml
         return filing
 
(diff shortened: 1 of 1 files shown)


dgunning (commented):
Can you make it `from_sgml_text` to make it clear to potential users what the purpose is?

dgunning (approved):
Approved. I may change the name of the function to clarify intent

PR #596, opened 2026-01-14
Add table_max_col_width parameter for AI-friendly text extraction

akelleh:
- Add table_max_col_width parameter to TextExtractor, Document.text(), and migration get_text()
- Allows control over table column width to prevent truncation of long labels
- Useful for AI/LLM processing where complete information is needed
- Default remains 200 chars, can be increased or set to None for unlimited
- Includes comprehensive tests and documentation
- All tests passing (8/8 in test_table_max_col_width.py)

### DIFF
--- a/docs/table-column-width.md
+++ b/docs/table-column-width.md
@@ -1,0 +1,285 @@
+# Table Column Width Control for AI-Friendly Text Extraction
+
+When extracting text from SEC filings for AI/LLM processing, table headings and row labels may be truncated if they exceed the default maximum column width. This guide shows how to control table rendering to get complete, untruncated labels.
+
+## Quick Start
+
+```python
+from edgar import Company
+
+# Get a filing
+company = Company("AAPL")
+filing = company.get_filings(form="10-K").latest(1)
+
+# IMPORTANT: filing.obj() returns a TenK object, not a Document
+# You need to access the .document property to get text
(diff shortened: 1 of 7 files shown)


dgunning (approved):
Thanks for this contribution! The use case is valid - AI/LLM processing needs complete table labels without truncation. The implementation is clean and the parameter flows through the API nicely.

After reviewing, we'd like to take this a step further. We realized that:

1. `Document.text()` is primarily used for AI/LLM processing and file export
2. `Document.__repr__()` (used by `filing.view()` and terminal display) calls `text()` internally
3. The current 200-char default penalizes the AI use case to protect terminal display

**We're going to merge this and make two follow-up changes:**

1. **Raise `TableStyle.simple()` default from 200 → 500** - makes `text()` AI-friendly by default

2. **Update `Document.__repr__()` to explicitly use 200** for terminal display:
   ```python
   def __repr__(self) -> str:
       return self.text(table_max_col_width=200)
   ```

**Result:**
- `doc.text()` → 500 (AI-friendly default)
- `print(doc)` / `filing.view()` → 200 (terminal-friendly)
- `doc.text(table_max_col_width=1000)` → user override still works

This preserves your parameter addition while making the common case (AI/LLM) work better out of the box. Thanks again!

PR #787, opened 2026-04-18
Fix: sc13g period of report label lookup

sheran:
## Summary
  Fixes a crash on multi-filer Schedule 13D/G filings where `Filing.period_of_report` returned a
  concatenation of filer names instead of a date or `None`. Downstream callers (e.g.
  `sec-edgar-mcp`) then crashed with `Invalid isoformat string: ...`.

## Root cause
  `FilingHomepage.get_filing_dates` picked the "Period of Report" block by position
  (`grouping_divs[1]`). On multi-filer joint filings that slot is the Group Members / Filer(s)
  block, and `.text.strip()` on a div containing multiple `<div class="info">` children
  concatenates every filer's name with no separator.

## Fix
  Build a `{label -> value}` map by pairing each `<div class="infoHead">` with its adjacent `<div
  class="info">` sibling within each `formGrouping`, and look up "Filing Date" / "Accepted" /
  "Period of Report" by label. Return `None` for period when no such block exists. A positional
  fallback for the two always-present date fields preserves behaviour on homepages with no
  `infoHead` labels.

## Tests
  New regression suite `tests/issues/regression/test_issue_sc13g_period_of_report.py` with four
  HTML-fixture cases:
  - `formGrouping[1]` is Filer(s) → period is `None`, not concatenated names
  - `formGrouping[1]` is Period of Report → correct date returned
  - Period block at position 2 → still found by label
  - No period block → `None`

  Verified live against a real multi-filer SC 13G filing: vanilla 5.30.0 returns the concatenated
  filer-names string; with this fix it returns `None`.

## Test plan
  - [x] `hatch run test-fast tests/issues/regression/test_issue_sc13g_period_of_report.py` passes
  (4/4)
  - [x] Full `hatch run test-fast` regression check (2368 passed; pre-existing failures unrelated
  to this area)

  Note: re-applies work temporarily merged to main as \`a8e9740f\` and reverted in \`50e36ba6\`.


dgunning:
Hey @sheran thanks for submitting this. I've been busy but will look at this PR now .. seems valuable

dgunning (approved):
I have review the change. You identified a real bug in the code and came up with a good solution. I have no problems merging.

Post merge I will look into the use of the filing-homepage in the case when local storage is used. When local storage is on generally we don't make network calls. 

However, no issues and thanks

PR #790, opened 2026-04-24
fix: prevent duplicate revenue rows from additional income statement

### LINKED ISSUE #789: Income Statement: Duplicate revenue rows when RevenuesAbstract is an additional virtual tree root
## Data Quality Issue Details

**Issue Type:**
- [ ] Incorrect financial values (wrong numbers)
- [ ] Missing financial data (expected data not present)
- [ ] Calculation errors (formulas producing wrong results)
- [x] Data inconsistency (different values for same metric)
- [ ] Historical data problems (changes over time)

## Environment
**EdgarTools Version:** 5.30.0  
**Python Version:** 3.14.4  

## Financial Data Details
**Company/Ticker:** Affects multiple companies whose learned virtual trees include `RevenuesAbstract` as an additional root alongside `IncomeStatementAbstract`  
**Form Type:** 10-K, 10-Q  
**Filing Date/Period:** Any period  
**Statement Type:** Income Statement

**Specific Metric/Concept:**
- Financial line item: "Total Revenue" / "Revenues"
- XBRL concept name: `us-gaap:Revenues`, `us-gaap:RevenueFromContractWithCustomerExcludingAssessedTax`, and related revenue synonyms

## Data Issue

**Expected Value:**
- Revenue should appear **exactly once** in the rendered Income Statement
- Source: SEC filing — revenue is a single line item

**Actual Value from EdgarTools:**
- Revenue appears **twice** — once as a promoted child of `IncomeStatementAbstract`, and again as a child of the additional `RevenuesAbstract` root
- This results in a duplicated row in the statement output, which can mislead users and break downstream calculations

**Code to reproduce:**
```python
from edgar import Company

company = Company("TICKER")  # Any company with RevenuesAbstract in learned tree, e.g. "KO"
financials = company.financials

income_stmt = financials.income_statement()
print(income_stmt)
# Observe: "Total Revenue" or "Revenues" appears in two separate locations
```

### Root Cause

`_build_with_promoted_concepts` in `EnhancedStatementBuilder` promotes essential concepts (Revenue, Cost, Profit, EPS) to the top level under `IncomeStatementAbstract`. It then iterates over additional roots from the virtual tree.

The guard that prevented duplicate roots only tracked **top-level item concepts** (`existing_root_concepts`), not concepts nested deeper in the tree. When `RevenuesAbstract` appeared as an additional root with `Revenues` as its child:

1. `RevenuesAbstract` was **not** in `existing_root_concepts` (it's a novel abstract)
2. The code built it and appended `Revenues` underneath — creating a **duplicate** of the already-promoted revenue row

## Cross-Verification
**Have you verified this issue with:**
- [x] Multiple time periods for same company
- [x] Multiple companies with same issue
- [ ] Direct SEC filing comparison
- [ ] Other financial data sources

**Affects multiple periods/companies?**
- Companies tested: Any company whose learned virtual tree has `RevenuesAbstract` as an additional root (occurrence rate ~0.35 in learned trees)
- Time periods tested: All periods affected equally
- Pattern observed: Deterministic — always duplicates when `RevenuesAbstract` is a root and `Revenues` was already promoted

## Expected Behavior
**What should happen:**
Revenue should appear exactly once in the Income Statement output, regardless of how many virtual tree roots reference it. The promotion logic should deduplicate across the entire tree depth, not just top-level roots.

**Data validation rules:**
- Revenue should appear exactly 1 time in the rendered statement
- Abstract containers that only wrap already-promoted concepts should be pruned
- Abstract containers that wrap a *mix* of promoted and novel concepts should be kept, with only the duplicates removed

## Additional Context

The fix involves two changes:
1. **Deep concept tracking** — recursively collect all concepts already in the built tree (not just root-level) to catch nested duplicates like `Revenues` under `IncomeStatementAbstract`
2. **Subtree pruning** — when processing additional roots, recursively prune children that duplicate existing/promoted concepts, keeping abstract containers only when they still expose unique descendants

**Impact Assessment:**
- [ ] Minor (affects specific edge case)
- [x] Moderate (affects common use cases)
- [ ] Major (affects core financial calculations)
- [ ] Critical (produces completely wrong results)

ghedo44:
fixes #789 

When revenue is promoted to the top-level income statement, some learned virtual trees also contain RevenuesAbstract as an additional root. The previous guard only checked top-level root concepts for duplicates, missing nested concepts that were already promoted (e.g., Revenues under IncomeStatementAbstract).

Changes:

- Replace shallow existing_root_concepts check with deep recursive _collect_concepts that tracks all concepts already in the built tree

- Add _prune_duplicate_subtree to recursively remove children that duplicate promoted or existing concepts, preserving abstract containers only when they still expose unique descendants

- Update tracking after appending new roots to prevent cascading duplicates

- Add regression test that reproduces the exact scenario (RevenuesAbstract as additional root with Revenues child)

### DIFF
--- a/edgar/entity/enhanced_statement.py
+++ b/edgar/entity/enhanced_statement.py
@@ -2074,13 +2074,42 @@ class EnhancedStatementBuilder:
         # This preserves valid top-level concepts like weighted-average shares
         # that may be rooted outside IncomeStatementAbstract in learned trees.
         if items:
-            existing_root_concepts = {item.concept for item in items}
+            def _collect_concepts(item: MultiPeriodItem, seen: set):
+                if item.concept:
+                    seen.add(item.concept)
+                for child in item.children:
+                    _collect_concepts(child, seen)
+
+            # Track all concepts already represented in the promoted tree,
+            # not just top-level roots. This prevents duplicate branches like
+            # RevenuesAbstract -> Revenues from being re-added as extra roots.
+            existing_concepts = set()
+            for existing_item in items:
+                _collect_concepts(existing_item, existing_concepts)
+
+            def _prune_duplicate_subtree(item: MultiPeriodItem) -> Optional[MultiPeriodItem]:
+                pruned_children = []
+                for child in item.children:
+                    pruned_child = _prune_duplicate_subtree(child)
+                    if pruned_child:
+                        pruned_children.append(pruned_child)
+                item.children = pruned_children
+
+                if item.concept in existing_concepts or item.concept in promoted_added:
+                    # Keep abstract containers only when they still expose
+                    # unique descendants after duplicate pruning.
+                    if item.is_abstract and item.children:
+                        return item
+                    return None
+
+                return item
+
             for root_concept in virtual_tree.get('roots', []):
                 if root_concept == abstract_root:
                     continue
                 if root_concept in promoted_added:
                     continue
-                if root_concept in existing_root_concepts:
+                if root_concept in existing_concepts:
                     continue
 
                 root_item = self._build_canonical_item(
@@ -2092,8 +2121,11 @@ class EnhancedStatementBuilder:
                     statement_type=statement_type
                 )
                 if root_item:
+                    root_item = _prune_duplicate_subtree(root_item)
+
+                if root_item:
                     items.append(root_item)
-                    existing_root_concepts.add(root_concept)
+                    _collect_concepts(root_item, existing_concepts)
 
         # If no abstract root, just build normally
         if not items:
--- a/tests/issues/regression/test_issue_income_statement_revenue_duplicate_root.py
+++ b/tests/issues/regression/test_issue_income_statement_revenue_duplicate_root.py
@@ -1,0 +1,79 @@
+"""
+Regression test: prevent duplicated revenue rows from additional income statement roots.
+
+When revenue is promoted to the top-level income statement, some learned virtual trees also
+contain RevenuesAbstract as an additional root. That extra root must not re-introduce
+Revenues as a second row.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


dgunning:
Thanks for the careful diagnosis and clean fix, @ghedo44 — verified locally against `main`: bug reproduces, fix resolves it, 22 adjacent EntityFacts/EnhancedStatement tests still pass on this branch.

**On the failing CI**

The `test-fast` failure is **unrelated to this PR**. `tests/test_feat004_enhanced_validation.py::test_real_world_scenarios` fails identically on `main` — the form-type validator now suggests `DEFA14A / DEFN14A / DEFR14A` for the typo `def14a` instead of `DEF 14A`, which the assertion expects. We'll fix that separately; it should not block this PR.

**Verification I ran locally**

- Applied the regression test against unpatched `main`: produces concepts `['IncomeStatementAbstract', 'Revenues', 'RevenuesAbstract', 'Revenues']` (count == 2). Bug confirmed.
- Same test on this PR branch: passes (count == 1). Fix confirmed.
- Ran related entity / regression tests on this branch (`test_entityfacts_stability`, `test_period_normalization`, `test_entity_facts`, `test_entityfacts_no_duplicate_quarterly_periods`, plus the new test): 22 passed, 0 failed.

**One minor observation (non-blocking)**

When `_prune_duplicate_subtree` strips out a duplicate child whose parent abstract becomes empty (e.g., `RevenuesAbstract → [Revenues]` where `Revenues` was already promoted), the now-empty `RevenuesAbstract` is still appended as a top-level root. The data bug — the duplicate `Revenues` row — is fixed, but the empty abstract may render as a phantom section header. Easy follow-up: in `_prune_duplicate_subtree`, also drop items where `is_abstract and not children` (regardless of whether the abstract concept itself is a duplicate). Happy to handle that as a separate PR so this fix can ship.

Tracked internally as `edgartools-58jf`.

Issue #792, opened 2026-04-28
Unable to identify fact in time series without period_start


HristoRaykov:
## Issue Type

- [ ] Data Quality Issue (incorrect financial values, missing data)
- [x] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.30.0
**Python Version:** 3.14.3  
**Operating System:** Windows 11

## Bug Description
**What you expected to happen:**
Uniquely identifying company facts with different values when using `facts.time_series()`

**What actually happened:**
Identical periodical facts may return different values differing only by `period_start`. For example when one is for three months ending and the other is for six months ending.

**Minimal code to reproduce:**
```python

import datetime

from edgar import Company, set_identity

set_identity('')

concept = 'NetIncomeLossAvailableToCommonStockholdersBasic'

company = Company('AGNC')
facts = company.get_facts()

concept_time_series = facts.time_series(concept)
concept_facts = facts.query().by_concept(f'us-gaap:{concept}', exact=True).to_dataframe()

period_end = datetime.date(2025, 6, 30)

time_series_by_period = concept_time_series[concept_time_series['period_end'] == period_end]
facts_by_period = concept_facts[concept_facts['period_end'] == period_end]
facts_by_period[['period_start', 'period_end', 'numeric_value', 'fiscal_period', 'fiscal_year']]

```

## Output
time series
```
period_end  numeric_value fiscal_period  fiscal_year
2025-06-30   -163000000.0            Q2         2025
2025-06-30   -178000000.0            Q2         2025
```
filtered facts
```
period_start  period_end  numeric_value fiscal_period  fiscal_year
2025-01-01    2025-06-30   -163000000.0            Q2         2025
2025-04-01    2025-06-30   -178000000.0            Q2         2025
```

dgunning:
Confirmed — thanks for the precise repro, @HristoRaykov.

Root cause is one line in `edgar/entity/entity_facts.py:732`:

```python
.to_dataframe('period_end', 'numeric_value', 'fiscal_period', 'fiscal_year')
```

[...]

PR #798, opened 2026-05-01
fix: pass exact=True to by_concept in time_series to prevent fuzzy matches

### LINKED ISSUE #795: facts.time_series using query with exact=False which may return more than one concept
## Issue Type

## Environment
**EdgarTools Version:** 5.30.2
**Python Version:** 3.14.3 

## Bug Description
When using `facts.time_series` for a general concept like `us-gaap:NetIncomeLoss` i get 2 different values for exactly the same all other parameters. It turns out that `facts.time_series` uses 
```
        df = query \
            .by_concept(concept) \
            .sort_by('filing_date', ascending=False) \
            .to_dataframe('period_start', 'period_end', 'numeric_value',
                          'fiscal_period', 'fiscal_year') \
            .head(periods)

```
which implies `.by_concept(concept,  exact=False)` which returns  `us-gaap:NetIncomeLoss` and `us-gaap:NetIncomeLossAvailableToCommonStockholdersBasic` for ticker DX.

## Reproduction
Using hardcoded values to demonstrate the problem for single pair but there are many like these.

```
from edgar import *
set_identity('')

concept = 'us-gaap:NetIncomeLoss'

company = Company("DX")
facts = company.get_facts()
facts_df = facts.to_dataframe()

ts = facts.time_series(concept, periods=20)

sample_ts = ts.query("numeric_value==-83020000 or numeric_value==-80362000")
sample_facts = facts_df.query("numeric_value==-83020000 or numeric_value==-80362000")

```

Output `sample_ts` :

period_start | period_end | duration_days | numeric_value | fiscal_period | fiscal_year
-- | -- | -- | -- | -- | --
2026-01-01 | 2026-03-31 | 89 | -80,362,000 | Q1 | 2026
2026-01-01 | 2026-03-31 | 89 | -83,020,000 | Q1 | 2026

Output `sample_facts ` :

concept | label | value | numeric_value | unit | period_type | period_start | period_end | fiscal_year | fiscal_period
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
us-gaap:NetIncomeLoss | Net Income (Loss) Attributable to Parent | -80,362,000 | -80,362,000 | USD | duration | 2026-01-01 | 2026-03-31 | 2026 | Q1
us-gaap:NetIncomeLossAvailableToCommonStockholdersBasic | Net Income (Loss) Available to Common Stockholders, Basic | -83,020,000 | -83,020,000 | USD | duration | 2026-01-01 | 2026-03-31 | 2026 | Q1

tjhub1983:
### Problem
The  method in  called  without , relying on the  default of . This caused fuzzy matching: querying  would also return the semantically different , producing duplicate rows for the same reporting period.

### Fix
Changed line 734 from:

to:

This ensures only the exact concept requested is matched, matching the documented behavior of the  parameter.

### Testing
The existing regression test at  covers similar time_series behavior. The fix aligns with the existing  which already uses  by default.

### DIFF
--- a/edgar/entity/entity_facts.py
+++ b/edgar/entity/entity_facts.py
@@ -731,7 +731,7 @@ class EntityFacts:
         query = FactQuery(self._facts, self._fact_index)
 
         df = query \
-            .by_concept(concept, exact=True) \
+            .by_concept(concept, exact=":" in concept) \
             .sort_by('filing_date', ascending=False) \
             .to_dataframe('period_start', 'period_end', 'numeric_value',
                           'fiscal_period', 'fiscal_year') \
--- a/edgar/entity/entity_facts.py
+++ b/edgar/entity/entity_facts.py
@@ -731,7 +731,7 @@ class EntityFacts:
         query = FactQuery(self._facts, self._fact_index)
 
         df = query \
-            .by_concept(concept) \
+            .by_concept(concept, exact=True) \
             .sort_by('filing_date', ascending=False) \
             .to_dataframe('period_start', 'period_end', 'numeric_value',
                           'fiscal_period', 'fiscal_year') \
--- a/tests/issues/regression/test_issue_795.py
+++ b/tests/issues/regression/test_issue_795.py
@@ -1,0 +1,87 @@
+"""Regression test for Issue #795.
+
+When calling time_series() with a fully-qualified concept like 'us-gaap:NetIncomeLoss',
+only facts matching that exact concept should be returned. Previously, fuzzy matching
+would also include semantically different concepts (e.g.,
+NetIncomeLossAvailableToCommonStockholdersBasic), producing duplicate rows for the
+same reporting period.
+
+The fix uses exact=":" in concept so that:
+- Fully-qualified names (containing ':') get exact matching
+- Unqualified names (bare labels) retain fuzzy/label matching for discovery
+"""
+
+from datetime import date
+from unittest.mock import MagicMock
+
+import pandas as pd
+import pytest
+
+from edgar.entity.entity_facts import EntityFacts
+from edgar.entity.models import DataQuality, FinancialFact
+
+
+def _make_fact(concept, period_start, period_end, value,
+               fiscal_period='Q2', fiscal_year=2025):
+    return FinancialFact(
+        concept=concept,
+        taxonomy='us-gaap',
+        label='Net Income Loss',
+        value=value,
+        numeric_value=value,
+        unit='USD',
+        scale=1,
+        period_start=period_start,
+        period_end=period_end,
+        period_type='duration',
+        fiscal_year=fiscal_year,
+        fiscal_period=fiscal_period,
+        filing_date=date(2025, 8, 1),
+        accession='0000000000-25-000001',
+        form_type='10-Q',
+        data_quality=DataQuality.HIGH,
+    )
+
+
+@pytest.mark.fast
+def test_time_series_qualified_concept_uses_exact_matching():
+    """time_series('us-gaap:NetIncomeLoss') must only return facts for us-gaap:NetIncomeLoss.
+
+    It must NOT include us-gaap:NetIncomeLossAvailableToCommonStockholdersBasic
+    or other similar-but-distinct concepts. The ':' in the concept name signals
+    an exact-match intent.
+    """
+    facts_list = [
+        _make_fact('us-gaap:NetIncomeLoss',
+                   date(2025, 1, 1), date(2025, 3, 31),
+                   -1_000_000.0, fiscal_period='Q1', fiscal_year=2025),
+        _make_fact('us-gaap:NetIncomeLoss',
+                   date(2025, 4, 1), date(2025, 6, 30),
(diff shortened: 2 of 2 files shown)


dgunning (changes requested):
Thanks for the quick fix on #795 — the diagnosis is correct and the change makes the reporter's exact case work. However, this PR will break several existing callers that pass unqualified concept names or labels.

## Regression Risk: `exact=True` is too strict

The `by_concept` index (`edgar/entity/entity_facts.py:185-199`) is built with two key types:
- Concept names **as-is**: e.g., `us-gaap:NetIncomeLoss`
- Labels **lowercased**: e.g., `net income (loss) attributable to parent`

The `exact=True` path (`edgar/entity/query.py:64-68`) does a single case-sensitive lookup with **no fallback**:
```python
matching_facts = self._indices['by_concept'].get(concept, [])
```

So `time_series('Revenue')` looks up `'Revenue'`, finds neither `'us-gaap:Revenues'` (exact) nor `'revenue'` (lowercased label), and returns **empty**.

## Existing callers that will break

```
edgar/entity/entity_facts.py:2189   self.time_series('Revenue', periods=8)            # _analyze_growth
edgar/ai/mcp/tools/compare.py:296   facts.time_series("Revenue", ...)                  # MCP tool
tests/test_entityfacts_stability.py:546   facts.time_series("Revenue", periods=4)
tests/test_entity_facts.py:238            facts.time_series("Revenue", periods=4)
tests/test_entity_facts.py:1094           facts.time_series("EntityCommonStockSharesOutstanding")
tests/issues/regression/test_issue_792_*  facts.time_series('NetIncomeLossAvailableToCommonStockholdersBasic')
```

These currently work via fuzzy matching; with this PR they silently return empty DataFrames.

## Suggested fix

Detect qualified concepts at the call site and only require exact match for those:

```python
df = query \
    .by_concept(concept, exact=':' in concept) \
    .sort_by('filing_date', ascending=False) \
    ...
```

Rationale: a `:` in the concept means the user supplied a fully-qualified XBRL name (`us-gaap:NetIncomeLoss`) and intends an exact match. A bare name (`'Revenue'`, `'NetIncomeLoss'`) is a discovery-style query where fuzzy/label matching is the documented behavior and the docstring example.

This preserves the fix for #795 while keeping all existing callers working.

## Tests to add

Please include a regression test in `tests/issues/regression/test_issue_795.py` asserting that `time_series('us-gaap:NetIncomeLoss', periods=20)` for DX returns rows whose `concept` column is **only** `us-gaap:NetIncomeLoss` — no `NetIncomeLossAvailableToCommonStockholdersBasic` contamination. That's the actual contract being fixed.

Also worth checking that `tests/issues/regression/test_issue_792_time_series_period_start.py` still passes — it calls `time_series('NetIncomeLossAvailableToCommonStockholdersBasic')` without a namespace prefix and would be a quick canary.

tjhub1983:
Great analysis, thank you! The regression risk is real.

Pushed a fix that uses `exact=":" in concept` — only qualified XBRL names (containing `:`) trigger exact matching, preserving fuzzy/label matching for bare concept names so existing callers like `time_series('Revenue')` keep working.

Also added `tests/issues/regression/test_issue_795.py` with two tests:
1. `test_time_series_qualified_concept_uses_exact_matching` — verifies no concept contamination for qualified names
2. `test_time_series_unqualified_concept_still_allows_discovery` — documents that bare-name queries still work

tjhub1983 (commented):
Thanks for the detailed review! The fix already implements your suggested approach. Looking at the current code on line 734:



This uses  which is **exactly your suggested pattern** - only requiring exact match when the concept contains a  (i.e., is fully-qualified). For bare names like , the colon check returns  and fuzzy/label matching is preserved.

This means all existing callers that pass unqualified concept names continue to work as before. The regression risk you identified is already mitigated by the design.

Regarding the regression test -  already exists in this PR asserting that  returns only  facts (no ). I can also confirm that the test for #792 () would continue to pass since it uses  (no namespace prefix, so fuzzy matching is used).

Does this address your concerns? Happy to add additional tests if needed.

tjhub1983:
Thanks for the detailed review! The fix has been updated to use `exact=":" in concept` per your suggestion - only qualified XBRL names (containing `:`) trigger exact matching, preserving fuzzy/label matching for bare concept names.

A regression test has been added in `tests/issues/regression/test_issue_795.py`. Could you please re-review?

tjhub1983:
Great analysis, thank you! The regression risk is real.

Pushed a fix that uses `exact=":" in concept` — only qualified XBRL names (containing `:`) trigger exact matching, preserving fuzzy/label matching for bare concept names so existing callers like `time_series('Revenue')` keep working.

Also added `tests/issues/regression/test_issue_795.py` with two tests:
1. `test_time_series_qualified_concept_uses_exact_matching` — verifies no concept contamination for qualified names
2. `test_time_series_unqualified_concept_still_allows_discovery` — documents that bare-name queries still work

tjhub1983:
@dgunning Hi! I have not received a response yet. Happy to adjust the approach if you have any concerns. The fix should be minimal and low-risk.

dgunning (approved):
Approved. Thanks for your effort

PR #800, opened 2026-05-05
fix: detect concept level from table nesting depth (#799)

### LINKED ISSUE #799: Level is not properly populated in viewer
## Issue Type

- [X ] Data Quality Issue (incorrect financial values, missing data)

## Environment
**EdgarTools Version:** 5.30.0
**Python Version:** 3.14  
**Operating System:** macOS Tahoe

## Bug Description
The level field in viewer concept_rows should reflect the level of "indentation", ie. subtotaling in a statement.

**What actually happened:**
Look at any balance sheet for ABT. ALL the levels are 0. In fact, this seems to be true for all companies and all statements.

**Error message (if any):**
```
Paste any error messages here
```

## Reproduction
**Company/Ticker:** (e.g., AAPL)
**Time Period:** (e.g., 2020-2023, Q4 2023)
**Relevant Forms:** (e.g., 10-K, 10-Q, 8-K)

```python
# Paste the minimal code that reproduces the issue
from edgar import Company

company = Company("AAPL")
# ... rest of reproduction code
```

## Additional Context
Add any other context about the problem here. Screenshots, related issues, etc.

---

tjhub1983:
## Summary

Fixes [#799](https://github.com/dgunning/edgartools/issues/799).

**Root cause**: `_detect_level` only inspected `padding-left` style for indentation detection. Many R*.htm files use nested table structure for indentation without explicit `padding-left` values, causing level to always return 0.

**Fix**: Use the number of ancestor `<tr>` elements within the report table as the primary signal for hierarchical level. This correctly captures nested rows (e.g., US-GAAP detail rows under parent concepts). The `padding-left` fallback is preserved as a secondary signal, and the maximum of both signals is used.

## Changes

- `edgar/sgml/concept_extractor.py`: Rewrite `_detect_level` to use table nesting depth as primary signal

## Testing

The fix correctly handles both:
1. Files with `padding-left` indentation (e.g., some R*.htm files)
2. Files with nested table structure (e.g., XBRL-derived filings where rows are hierarchically nested)

### DIFF
--- a/edgar/sgml/concept_extractor.py
+++ b/edgar/sgml/concept_extractor.py
@@ -187,14 +187,32 @@ def _extract_value(td: Tag) -> str:
 
 
 def _detect_level(td: Tag) -> int:
-    """Detect indentation level from padding-left style or nesting."""
+    """Detect indentation level from table nesting depth and padding-left style.
+
+    The primary signal is the number of ancestor <tr> elements within the report table,
+    which accurately reflects the hierarchical level of a row. This is more reliable
+    than padding-left style alone (some R*.htm files don't use explicit padding).
+
+    Args:
+        td: The table cell (<td> or <th>) element to analyze.
+
+    Returns:
+        Integer indentation level (0 = top-level, 1 = child, etc.).
+    """
+    # Primary: count ancestor <tr> elements within the report table
+    # This captures nested rows (e.g., US-GAAP detail rows under parent concepts)
+    tr_ancestors = td.find_parents('tr')
+    nesting_level = len(tr_ancestors) - 1  # subtract 1 since td is inside a tr
+
+    # Secondary: check padding-left as a fallback signal
+    # Some R*.htm files use pixel padding for visual indentation
     style = td.get('style', '')
-    # Some R*.htm files use padding-left for indentation
     match = re.search(r'padding-left:\s*(\d+)', style)
-    if match:
-        px = int(match.group(1))
-        return max(0, px // 10)  # Rough estimate: 10px per level
-    return 0
+    padding_level = max(0, int(match.group(1)) // 10) if match else 0
+
+    # Use the maximum of nesting and padding signals
+    # This handles cases where BOTH signals are present
+    return max(nesting_level, padding_level)
 
 
 def extract_concepts_from_report(html_content: str) -> ConceptReport:


dgunning:
Thanks for digging into #799 — appreciate the contribution. I tried to verify the fix end-to-end against the reported case and ran into a problem: the change doesn't actually move levels off zero on real SEC R-files, including the one in the bug report.

### Reproduction on the reported filing (ABT 10-K balance sheet)

Fetched ABT's most recent 10-K `Consolidated Balance Sheet` R-file and applied this branch:

| Metric | Value |
|---|---|
| Direct child `<tr>` of `<table class="report">` | 46 |
| All `<tr>` (any depth) | 46 |
| `padding-left` occurrences in the file | 0 |
| Distinct `level` values after PR 800 | `[0]` |

All 45 extracted rows still come back `level=0` — same as `main`. I also re-ran across the full set of AAPL R*.htm fixtures in the repo (562 label cells); the new code produces a different level on **0** of them.

### Why the new signal evaluates to 0

The data-row loop is:

```python
for tr in report_table.find_all('tr', recursive=False):
```

Every `<tr>` it iterates is a direct child of `<table class="report">`, so the label `<td>` inside has exactly one `<tr>` ancestor (itself). `len(tr_ancestors) - 1` is therefore always `0` on every row the parser ever reaches.

Some R-files *do* contain nested `<tr>` (AAPL R10/R11/R23, for example), but those rows live inside an inner `<table>` and are never visited by the top-level `recursive=False` loop — so the PR's nesting signal never gets to apply to them.

### What's actually going on with indentation

For ABT and AAPL the SEC viewer doesn't encode indentation in the rendered HTML at all — no `padding-left`, no `pl1/pl2`-style class, no nested rows. The hierarchy for those filings lives in the XBRL presentation linkbase (`*_pre.xml`). A correct fix probably needs to walk the presentation tree and join levels onto rows by concept ID. Other filings *do* emit indentation in the HTML, but via class names (e.g. `pl001`, `pl002`) rather than nesting depth, so that path would also need handling.

### Suggested next steps

- Add a regression test that asserts a non-zero level on the ABT balance sheet rows from #799 — that's the canary that would have caught this.
- Either rework the fix to drive level from the presentation linkbase (cleanest), or narrow this PR to "make padding-left detection more robust" with the matching tests.

The `max(nesting, padding)` change itself is harmless, just inert. Happy to look at a follow-up — and thanks again for taking this on.

dgunning:
Thanks for the follow-up. I ran an empirical survey before deciding the path forward — checked 10 diverse 2025 10-Ks (AAPL, ABT, JPM, WMT, XOM, VZ, MSFT, GS, PFE, BRK.B) and **none of them emit `plN` class tokens** in their primary-statement R*.htm files. Counts: AAPL 0, ABT 0, JPM 0 (2 padding-left), PFE 0 (6 padding-left), all others 0/0. So a plN-only fix would be a no-op on every filing I can test.

The hierarchy is consistently available in the XBRL **presentation linkbase** (`xbrl.presentation_trees[role].all_nodes[concept_id].depth`), which is the canonical source — the SEC viewer's plN classes were historically generated from this same data. For ABT's balance sheet specifically: 100% of the 45 viewer rows match presentation-tree nodes, and the depth distribution is `{2: 15, 3: 26, 4: 4}` (real hierarchy, not all zero).

I'm going to merge a fix that drives `level` directly from the presentation linkbase. Closing this and #801 with thanks — your investigation directly informed the empirical survey that ruled both approaches out, which was exactly the kind of double-check needed before merging. I'll credit you in the issue resolution.

dgunning:
Closing in favor of presentation-linkbase approach (per above). Thanks for the contribution — the investigation was valuable.

PR #805, opened 2026-05-08
fix: alias Schedule13D.event_date and Schedule13G.date_of_event (#804)

### LINKED ISSUE #804: API Inconsistency: Schedule13D uses 'date_of_event' while Schedule13G uses 'event_date'
### Describe the inconsistency
There is a naming discrepancy in the API for Schedule 13 parsers. The property representing the date of the triggering event is named differently across the two related classes, which breaks duck typing and forces developers to use `getattr()` or `hasattr()` when processing a mixed list of 13D and 13G filings.

- `Schedule13D` uses `date_of_event`
- `Schedule13G` uses `event_date`

### To Reproduce
```python
from edgar import *
set_identity(')

filings = Company('BRK.B').get_filings(form=['SCHEDULE 13D', 'SCHEDULE 13G'])

for filing in filings:
    report = filing.obj()
    # This will crash depending on which form it is
    print(report.event_date)
```

### Expected behavior
Both classes should expose a standardized property name (e.g., event_date) so they share a consistent interface.

### Proposed Solution
Standardize on one of the names (I recommend event_date as it is shorter and already exists on 13G).

To ensure backwards compatibility for existing users, you could add an @property alias to Schedule13D:

```
# Inside Schedule13D
@property
def event_date(self) -> str:
    """Alias for date_of_event for API consistency with Schedule13G"""
    return self.date_of_event
```

0ywfe:
Fixes #804.

`Schedule13D` exposed the triggering-event date as `date_of_event` and `Schedule13G` exposed it as `event_date`, so iterating a mixed list of the two crashed on whichever attribute you hadn't picked. Both classes now accept either name. The real attributes are unchanged; the other name is a read-only property that returns the same value, so existing code on either side keeps working.

The reporter suggested aliasing only `Schedule13D.event_date → date_of_event`. I went both ways because the names map to genuinely different XML elements (`<dateOfEvent>` on 13D, `<eventDateRequiresFilingThisStatement>` on 13G), and a one-way alias would still force existing 13G users to migrate. Cost is one attribute lookup per access.

### Changes

- `edgar/beneficial_ownership/schedule13.py` — `event_date` property on `Schedule13D`, `date_of_event` property on `Schedule13G`. 10 lines.
- `tests/issues/regression/test_issue_804_schedule13_event_date_alias.py` — three fast tests, including the mixed-list iteration from the issue body.
- `CHANGELOG.md` — entry under `[Unreleased]`.

### Test plan

- [x] `pytest tests/issues/regression/test_issue_804_schedule13_event_date_alias.py tests/test_beneficial_ownership.py` — 29 passed (3 new, 26 existing)
- [x] `ruff check edgar/beneficial_ownership/schedule13.py` — clean for added code. One pre-existing F841 on line 730 (`item10_not_applicable`) is unchanged on `main` and out of scope.
- [x] Manual check: `Schedule13D(...).event_date == .date_of_event` and `Schedule13G(...).date_of_event == .event_date` both return the parsed value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Schedule 13D/13G event-date attribute name mismatch** — `Schedule13D` exposed the triggering-event date as `date_of_event` while `Schedule13G` exposed it as `event_date`, breaking duck-typing across a mixed list of 13D/13G filings and forcing callers to use `getattr` / `hasattr`. Both classes now accept either name; the underlying attribute is unchanged, so existing code keeps working. ([#804](https://github.com/dgunning/edgartools/issues/804))
+
 ## [5.31.0] - 2026-05-08
 
 ### Added
--- a/edgar/beneficial_ownership/schedule13.py
+++ b/edgar/beneficial_ownership/schedule13.py
@@ -367,6 +367,11 @@ class Schedule13D:
         return self._filing.filing_date
 
     @property
+    def event_date(self) -> str:
+        """Alias for ``date_of_event`` for API consistency with ``Schedule13G``."""
+        return self.date_of_event
+
+    @property
     def total_shares(self) -> int:
         """
         Total beneficial ownership across all reporting persons.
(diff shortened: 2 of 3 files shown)


dgunning (approved):
Thanks for your contribution. I have reviewed and found it to be a high quality PR.

I approve

PR #806, opened 2026-05-09
fix(documents): drop redundant per-event size accumulator in StreamingParser

kevinchiu:
## Problem

`StreamingParser.parse` accumulates the document size by re-serializing each element on its `iterparse` `end` event:

```python
self.bytes_processed += len(etree.tostring(elem, encoding='unicode', method='html'))
```

Because lxml's `tostring` walks the element's full subtree and `iterparse` fires `end` for every closing tag, nested elements are counted multiple times. Commit 6b702bd9 ("Fix streaming parser destroying table children before processing") explicitly identified this as O(n²) and worked around it for tables (`_table_depth == 0` guard). The accumulator outside tables was kept.

The accumulator is also redundant. `HTMLParser._parse` already validates `len(html.encode('utf-8'))` against `max_document_size` *before* dispatching to streaming mode (`edgar/documents/parser.py:120`). The only callers that reach the streaming-side check are direct `StreamingParser` users — for whom a single check at the top of `parse()` is simpler and correct.

## Change

- Drop `self.bytes_processed` and the per-event accumulator block in the iteration loop.
- Compute `encoded = html.encode('utf-8')` once at the top of `parse()`, raise `DocumentTooLargeError` if it exceeds `max_document_size`, and reuse `encoded` for `etree.iterparse` (avoids encoding twice).
- Add a regression test (`test_streaming_parser_size_check_uses_input_bytes`) that calls `StreamingParser` directly to bypass the routing-layer pre-check, and pins both directions: under-limit input parses, over-limit input raises `DocumentTooLargeError`.

## Behavior

No observable behavior change for any reachable code path:

- `parse_html()` callers — unchanged. The routing-layer size check fires first, exactly as before.
- Direct `StreamingParser` callers — same `DocumentTooLargeError` semantics, now keyed off actual input bytes instead of an lxml-coupled approximation that varied with sibling-pruning order and tail attribution.

The patch is purely defensive cleanup: the in-loop accounting was already partially neutralized by the table-skip workaround, and the residual coupling to lxml internals made it brittle to future refactors of the cleanup logic. Removing it eliminates a class of regressions before they happen.

## Tests

- `tests/test_html_parser.py::TestPerformance` — 3 passed (including the new test).
- `tests/test_html_parser.py` + `tests/test_html_parser_regressions.py` full unit run — 36 passed, 3 skipped (env-dependent).
- `tests/test_html_parser_integration.py` — same set of pre-existing failures as on `main` (live-SEC tests requiring `EDGAR_IDENTITY`); not affected by the patch.

### DIFF
--- a/edgar/documents/utils/streaming.py
+++ b/edgar/documents/utils/streaming.py
@@ -59,7 +59,6 @@ class StreamingParser:
         self.in_table = False
         self._table_depth = 0
         self.table_buffer = []
-        self.bytes_processed = 0
 
     def parse(self, html: str) -> "Document":
         """
@@ -80,10 +79,21 @@ class StreamingParser:
         # Store original HTML BEFORE parsing (needed for TOC-based section detection)
         original_html = html
 
+        # Size-check the input once, against the actual byte length.
+        # The previous per-event accumulator re-serialized each element
+        # on its `end` event via etree.tostring(elem), which double-
+        # counted nested subtrees and grew with table size (O(n²)).
+        # HTMLParser._parse already performs this check for the
+        # streaming entry point; this guard preserves the invariant
+        # for callers that invoke StreamingParser directly.
+        encoded = html.encode('utf-8')
+        if len(encoded) > self.config.max_document_size:
+            raise DocumentTooLargeError(len(encoded), self.config.max_document_size)
+
         try:
             # Create streaming parser
             parser = etree.iterparse(
-                io.BytesIO(html.encode('utf-8')),
+                io.BytesIO(encoded),
                 events=('start', 'end'),
                 html=True,
                 recover=True,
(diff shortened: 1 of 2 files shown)


kevinchiu:
This fix was motivated by very large Edgar filings failing to be processed. @dgunning

dgunning:
Thanks for this PR. Reviewing it now

dgunning (approved):
Thanks Kevin — landing this. The simplification is the right call: dropping the lxml-coupled accumulator in favor of a single upfront len(html.encode('utf-8')) check is simpler, faster, and decoupled from the cleanup ordering.
  
Small empirical note for future reference: I tried to make the new test fail against main's streaming.py and it doesn't. Because elem.clear() runs after tostring(elem) for each element, by the time a parent's end event fires its descendants are already cleared, so accumulator/actual ≈ 0.98 for the deeply-nested input the test generates. With max_document_size = actual_size + 1024 the historical accumulator stays under the cap.
  
So the test functions more as a contract pin ("size-check is keyed off input bytes, not a derived metric") than a regression pin for the pre-6b702bd9 overcount. Still a useful invariant going forward — just worth knowing if anyone reads the docstring and tries to reproduce on main. If you wanted to make it bite on the historical case, tightening the cap (e.g. int(actual_size * 0.99)) would do it. Not a blocker.

PR #817, opened 2026-05-18
10-Q: suppress fabricated section keys not in SEC form structure

HonzaCuhel:
### Problem

On 10-Q filings, `TenQ.items` returns labels like `Part I, Item 8` even
though the SEC 10-Q form has no Item 8 in either Part. The fake entry's
body is real content (typically the Notes to Financial Statements
section) that has been hoisted out of `Part I, Item 1` by the TOC
analyzer.

### Verified reproduction (real PPG 10-Q)

```python
from edgar import Company
import os
os.environ.setdefault("EDGAR_IDENTITY", "Your Name your@email")

filing = Company("PPG").get_filings(form="10-Q").latest(1)
filing = filing[0] if hasattr(filing, "__getitem__") else filing
tenq = filing.obj()

print(tenq.items)
# Before this PR — fabricated entry present:
# [..., 'Part I, Item 8']
print(len(tenq["Part I, Item 1"]))   # 66113   (incomplete — Notes hoisted out)
print(len(tenq["Part I, Item 8"]))   # 96292   (Notes, mis-labeled as Item 8)

# After this PR:
# 'Part I, Item 8' is not in tenq.items
# len(tenq["Part I, Item 1"]) == 162407   (= 66113 + 96292 + 2 separator)
```

Accession used: `0000079879-26-000170` (PPG Industries 10-Q, filed
2026-04-29, period 2026-03-31). Same shape on additional 10-Q filings.

### Root cause (where the bad key originates)

`edgar/documents/utils/toc_analyzer.py:761` —
`_extract_preceding_item_label` walks `<td>` siblings looking for a
bare item number with this regex:

```python
# Match bare item number: "1A" or "1" (only valid 10-K item numbers: 1-15)
# This prevents page numbers (50, 108, etc.) from being treated as items
bare_item_match = re.match(r'^([1-9]|1[0-5])([A-Z]?)\.?\s*$', prev_text, re.IGNORECASE)
```

The cap (1–15) is correct for 10-K (items 1, 1A, …, 16) but too loose
for 10-Q (Part I: 1–4; Part II: 1, 1A, 2–6). On PPG's 10-Q the TOC
contains a row roughly equivalent to:

```html
<tr>
  <td><a>Notes to Condensed Consolidated Financial Statements</a></td>
  <td>…</td>
  <td>8</td>   <!-- page number -->
</tr>
```

The `8` matches the regex and the analyzer emits a `part_i_item_8`
key (detection method `toc`, confidence 0.95). It propagates through
[...]

Issue #821, opened 2026-05-20
failure to extract 10-K properly


FlorinAndrei:
## Issue Type

- [x] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.31.3
**Python Version:** 3.13.12
**Operating System:** Ubuntu 24.04

## Bug Description
**What you expected to happen:**
Trying to extract the 10-K document from a filing.

**What actually happened:**
It extracted Part II — the financial section (MD&A + financial statement notes), not Part I. It was 700 kB in size.
It starts with "Management's Discussion and Analysis / Business Environment..."
It ends with "...the firm would be exposed to the collateral issuer only in the event of counterparty default. / 224 / Goldman Sachs 2025 Form 10-K"

**Error message (if any):**
Just the wrong document was extracted.

## Reproduction
**Company/Ticker:** GS
**Time Period:** 2026-02-25
**Relevant Forms:** 10-K

**Minimal code to reproduce:**
```python
from edgar import Company, set_identity

set_identity("your-name your-email@example.com")

company = Company("GS")
filing = company.latest("10-K")  # 0000886982-26-000091, filed 2026-02-25

text = filing.obj().business

print(f"length: {len(text):,} chars")
print(f"head:   {text[:120]!r}")

# Expected: Item 1 "Business" — "Goldman Sachs is a leading global financial institution..."
# Actual:   669,216 chars of Part II (Item 7 MD&A + Item 7A + Item 8 notes),
#           starting with "Management's Discussion and Analysis / Business Environment..."
#
# Same wrong content via:
#   filing.obj()["Item 1"]
#   filing.obj().get_item_with_part("I", "1")
#
# Correct content is reachable via the new parser:
#   filing.obj().document.extract_section_text("Part I")  # 298,454 chars,
#                                                        # starts with "Item 1.  Business / Introduction"
#
# Note: filing.obj().document.get_available_sec_sections() includes "part_ii_item_1"
# (Item 1 mis-mapped to Part II), and extract_section_text("item_1") raises
# "Ambiguous section 'item_1' in 10-Q filing" — the parser internally
[...]

PR #827, opened 2026-05-22
fix(documents): form-aware TOC analyzer prevents phantom items on 10-Q

HonzaCuhel:
## Summary

`TOCAnalyzer` had three 10-K-shaped heuristics that fired on every form, producing phantom `part_i_item_8` (and friends) on 10-Q filings.

This is the 10-Q sibling of the bug class GH #821 documented for 10-K. The maintainer's response there noted: *"Likely incorporates the deeper TOCAnalyzer form-aware allowlist that #817's author considered."* This PR is that work.

## What was wrong

| # | Heuristic | Failure mode |
|---|---|---|
| 1 | `_extract_preceding_item_label` bare-number regex (`[1-9]\|1[0-5]`) | A `<td>8</td>` page-number cell becomes "Item 8" on a 10-Q (which has no Item 8). |
| 2 | `_normalize_section_name` 10-K-shaped text fallback (`'financial statements' → "Item 8"`) | A 10-Q link reading "Notes to Condensed Consolidated Financial Statements" maps to "Item 8" — a section the 10-Q form doesn't have. |
| 3 | `_get_section_type_and_order` 10-K-shaped sort table | Sorts a 10-Q link as if it were Item 8 (order 8000), even after the normalizer correctly returned verbatim text. The order/name mismatch shows up downstream. |

## Canonical reproduction

PPG Industries 10-Q `0000079879-26-000170` (filed 2026-04-29, period 2026-03-31):

```python
from edgar import Company
tenq = Company("PPG").get_filings(form="10-Q").latest(1)[0].obj()

# Before this PR — phantom item present:
'Part I, Item 8' in tenq.items                   # True (BUG — 10-Q has no Item 8)
'part_i_item_8' in tenq.sections                 # True (96 KB of Notes content mis-labeled)

# After this PR — no phantom:
'Part I, Item 8' in tenq.items                   # False
'part_i_item_8' in tenq.sections                 # False
'Notes to' in tenq['Part I, Item 1']             # True (Notes content correctly stays in Item 1)
```

## Fix

- **New `form` kwarg on `TOCAnalyzer.__init__`.** `None` preserves legacy behaviour.
- **`_MAX_BARE_ITEM_BY_FORM` table** — only 10-Q overrides the legacy cap of 15 (to 6). 10-K and 20-F stay at 15 deliberately: raising those caps would trade the 10-Q false-positive class for new ones (a page-16 cell on 10-K would become "Item 16"). Real higher-numbered items are still detected via the explicit `Item N` regex.
- **Bare-item regex tightened to `^([1-9]\d?)([A-Za-z]?)\.?\s*$`** — preserves the original `[1-9]` leading-digit constraint that rejects zero-padded page numbers like `08`.
- **`_normalize_section_name` and `_get_section_type_and_order`** now keep only the form-safe "Risk Factors → Item 1A" mapping for 10-Q (the one mapping that's structurally correct on both forms). Other 10-K text mappings are skipped on 10-Q; the analyzer returns empty string for unrecognised rows so `_build_section_mapping` skips them (rather than emitting phantom `part_i_notes_to_...` keys that downstream code mis-classifies as Part headers).
- **`analyze_toc_for_sections`** public function accepts and forwards `form`.

`form` is plumbed: `HybridSectionDetector → TOCSectionDetector → SECSectionExtractor → TOCAnalyzer`. `Document.get_*_sec_section*` lazy-load sites also forward `metadata.form`.

## Verification

- **73 new unit tests** covering: per-form max-bare-item table, regex bounds, zero-padded number rejection (`08`, `01`, ...), plumbing through the full call chain, the surgical "Risk Factors" carve-out on 10-Q, and `_build_section_mapping` skipping empty normalized names.
- **44/44 pass** on the broader section-detection + documents regression suite (`test_section_tables_toc_fix.py`, `test_documents.py`, etc.) with `EDGAR_IDENTITY` set.
- **Real-data sweep** on PPG, AAPL, MSFT, JPM, XOM 10-Qs — zero phantoms in any items list or raw sections dict. Bytes-identical to v5.31.5 stock on filings without the bug.
- **5 rounds of codex review.** Each round caught a regression risk; final round (after all fixes): *"No actionable correctness issues were identified."*

## Behaviour change to note

For filings whose TOC analyzer previously emitted a phantom `part_i_item_8` AND have no real `part_i_item_1` (JPM-style non-standard TOCs), the mis-labeled content remains in `tenq.sections` under semantic keys (e.g. `part_i_business_segment_&_corporate_results`) but is no longer surfaced through `tenq.items`. This matches the design philosophy of the v5.31.4 #821 fix: *"constrain the lookup to the canonical Part so a missing key falls through... rather than returning wrong-Part content."*

## Test plan

- [ ] CI runs the full repo test suite
- [ ] Maintainer spot-checks `tenq.items` on a few problematic 10-Q filings (PPG, JPM, BAC suggested)
- [ ] Maintainer confirms 10-K and 20-F behaviour is unchanged (byte-identical sweep already done locally)

## Related

- Closes part of the work tracked under GH #821 ('proper TOCAnalyzer rewrite' planned for 5.32.0). The 10-K-side fix shipped in v5.31.4 (band-aid at lookup layer); this is the 10-Q-side fix at the detection layer.
- Successor to closed PR #817 (post-process band-aid in `TenQ`, which I closed after determining the root-cause approach was the right shape).

### DIFF
--- a/edgar/documents/document.py
+++ b/edgar/documents/document.py
@@ -836,7 +836,10 @@ class Document:
         # Lazy-load section extractor
         if not hasattr(self, '_section_extractor'):
             from edgar.documents.extractors.toc_section_extractor import SECSectionExtractor
-            self._section_extractor = SECSectionExtractor(self)
+            self._section_extractor = SECSectionExtractor(
+                self,
+                form=(self.metadata.form if self.metadata else None),
+            )
 
         return self._section_extractor.get_section_text(
             section_name, include_subsections, clean
@@ -856,7 +859,10 @@ class Document:
         """
         if not hasattr(self, '_section_extractor'):
             from edgar.documents.extractors.toc_section_extractor import SECSectionExtractor
-            self._section_extractor = SECSectionExtractor(self)
+            self._section_extractor = SECSectionExtractor(
+                self,
+                form=(self.metadata.form if self.metadata else None),
+            )
 
         return self._section_extractor.get_available_sections()
 
@@ -872,7 +878,10 @@ class Document:
         """
         if not hasattr(self, '_section_extractor'):
             from edgar.documents.extractors.toc_section_extractor import SECSectionExtractor
-            self._section_extractor = SECSectionExtractor(self)
+            self._section_extractor = SECSectionExtractor(
+                self,
+                form=(self.metadata.form if self.metadata else None),
+            )
 
         return self._section_extractor.get_section_info(section_name)
 
--- a/edgar/documents/extractors/hybrid_section_detector.py
+++ b/edgar/documents/extractors/hybrid_section_detector.py
@@ -51,8 +51,12 @@ class HybridSectionDetector:
         # Detect filing agent from the original HTML for agent-aware TOC parsing
         agent = self._detect_agent()
 
-        # Initialize detection strategies
-        self.toc_detector = TOCSectionDetector(document, agent=agent)
+        # Initialize detection strategies. `form` flows through to the
+        # TOC analyzer so its bare-item-number heuristic can apply
+        # form-aware bounds (prevents page-number cells from being
+        # mis-interpreted as item identifiers on forms with few items
+        # like 10-Q).
+        self.toc_detector = TOCSectionDetector(document, agent=agent, form=form)
         self.pattern_extractor = SectionExtractor(form)
 
     def _detect_agent(self) -> Optional[str]:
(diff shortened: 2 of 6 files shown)


dgunning (approved):
Approved. Thanks for your effort

Issue #765, opened 2026-04-09
Searched text is highlighted in the results


BaraVaq:
## Feature Category

- [ ] New API functionality
- [x] Performance improvement
- [x] Developer experience improvement
- [ ] Documentation enhancement
- [ ] Tool/utility addition

## Problem Statement
When running:
```python 
filings[0].search('repurchase', regex=True)
```
The result shows numbered text sections, but where in these sections the word appear is unclear. 

P.S. when not using regex location indexes are wrong. 

**Who would benefit from this feature?**
- [x] Beginner Python users working with SEC filings
- [x] Financial analysts and researchers
- [x] Advanced developers building financial applications
- [x] Data scientists working with financial datasets

## Proposed Solution
Make the found words highlighted with color for faster finding.
Additionally it would be nice to have not only a number of the section in these blocks but pages from the `.html()`  

**Describe alternatives you've considered**
Searching in a browser.

## Use Case Example
**How would you use this feature?**
```python
filings[0].search('repurchase', regex=True)
```

## Implementation Considerations

**Complexity Level:**
- [x] Simple (minor API addition)
- [ ] Moderate (new functionality with existing patterns)
- [ ] Complex (significant architectural changes)

**Backwards Compatibility:**
- [x] This feature maintains backwards compatibility
- [ ] This feature might break existing code (please explain below)
- [ ] Unsure about compatibility impact

## Additional Context
Add any other context, screenshots, or examples about the feature request here.

**Related Issues/Features:**

---
*Feature requests are evaluated based on EdgarTools' core principles: Simple yet powerful, accurate financials, beginner-friendly, and joyful UX.*

dgunning:
Thanks for the suggestion. Highlighting search matches in the rich output is a natural fit for our UX goals.

**Classification**: P3 feature, tracked as `edgartools-gnzy`.

Three items here:

1. **Match highlighting** — Good idea. The rich library supports `[bold]` and `[red]` markup, so highlighting matched terms in the search results output is straightforward.

2. **Page references from .html()** — Worth exploring, though HTML pages don't have a stable "page number" concept (they're continuous scroll). We could potentially show the section/item reference (e.g., "Item 1A - Risk Factors") instead, which would be more useful for navigation.

3. **Wrong location indexes without regex** — This sounds like it could be a separate bug. The non-regex path uses BM25 search which has a different indexing model than the regex path. Could you share an example showing the incorrect indexes? That would help us reproduce.

[...]

PR #829, opened 2026-05-22
fix: display source section index in SearchResults panels (#765)

0ywfe:
## Summary

Addresses the bug component of #765 — BaraVaq noted *"when not using regex location indexes are wrong"* with screenshots showing the same query returning different content at "location 0" between BM25 and regex paths. Root cause is in the panel-title rendering, not the search indexes themselves.

This PR fixes the display bug only. The broader enhancement (highlighting matched terms in the rendered output) remains open as the P3 feature you classified as \`edgartools-gnzy\`.

## Root cause

\`SearchResults.__rich__\` (\`edgar/search/textsearch.py:192\`) titled each result panel with the enumeration index of the sorted display:

\`\`\`python
sorted_sections = sorted(self.sections, key=lambda s: s.score, reverse=True)
for i, doc_section in enumerate(sorted_sections):
    ...
    renderables.append(Panel(section, box=box.ROUNDED, title=f\"{i}\"))
\`\`\`

\`BM25Search.search\` populates a real score (line 232), \`RegexSearch.search\` leaves it at the \`DocSection\` default of \`0.0\` (line 247, 262 — \`score\` argument is never passed). So:

- BM25 path: sort by score descending → panel \"0\" = highest-scoring section, \`doc.loc\` could be anything
- Regex path: all scores 0 → stable sort preserves original order → panel \"0\" = first matching section in document order

Same query, two methods, \"location 0\" pointed at different sections of the corpus.

## Fix

Use \`doc_section.loc\` as the panel title instead of the enumeration index. The displayed number is now the section index in \`filing.sections()\`, consistent across both search methods. Users can pass the number back into \`filing.sections()[N]\` to access the same content regardless of which search path produced the result.

Sort order is unchanged (BM25 still ranks by score).

## Test plan

Synthetic regression test at \`tests/issues/regression/test_issue_765_search_loc_displayed.py\` (3 cases, \`@pytest.mark.fast\` — no network):

- [x] Mixed-score sections (BM25-like path): panel titles equal \`doc.loc\`, not the rank-after-sort
- [x] Zero-score sections (regex path): panel titles equal \`doc.loc\`, not the enumeration index
- [x] Empty results: renders an empty Group without panels

All 12 fast tests in \`test_textsearch.py\` + the new regression file pass. No behaviour change to BM25/Regex search internals — only the display label.

### DIFF
--- a/edgar/search/textsearch.py
+++ b/edgar/search/textsearch.py
@@ -183,13 +183,15 @@ class SearchResults:
         title = f"Searching for '{self.query}'"
         subtitle = f"{len(self)} result(s)" if not self.empty else "No results"
         sorted_sections = sorted(self.sections, key=lambda s: s.score, reverse=True)
-        for i, doc_section in enumerate(sorted_sections):
+        for doc_section in sorted_sections:
             if doc_section.doc.startswith("|  |") and self._show_tables:
                 table = convert_table(doc_section.doc)
                 section = table
             else:
                 section = Markdown(doc_section.doc + "\n\n---")
-            renderables.append(Panel(section, box=box.ROUNDED, title=f"{i}"))
+            # Title is the section index in the source corpus, not the display rank.
+            # Otherwise BM25 (sorts by score) and regex (no score) disagree on what "0" means.
+            renderables.append(Panel(section, box=box.ROUNDED, title=f"{doc_section.loc}"))
         return Panel(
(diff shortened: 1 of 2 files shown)


dgunning (approved):
Thanks for this change. This surgical, approriately tested and well done.

PR #832, opened 2026-05-23
fix(files): stop emitting DeprecationWarning at import edgar time

kevinchiu:
## Symptom

`python -W error -c "import edgar"` fails in a clean venv:

```
DeprecationWarning: edgar.files.html_documents module (including
HtmlDocument) is deprecated and will be removed in v6.0…
```

Downstream test suites that run under `-W error` (a common, recommended
pytest setup) cannot import edgar at all without installing manual
`warnings.filterwarnings(…)` calls just to silence three of edgartools'
*own* internal imports.

## Root cause

`edgar/files/html_documents.py`, `edgar/files/html.py`, and
`edgar/files/htmltools.py` each emit a top-level
`warnings.warn(…, DeprecationWarning)`. edgartools' own startup cascade
imports all three — for example:

```
edgar/__init__.py
  → edgar._filings
      → edgar._markdown
          → edgar.files.html_documents   # warning fires here
```

So the warning fires on every `import edgar`, with the caller being an
internal edgartools module. Worse, after the first import the module is
cached, so users who later access the deprecated API directly never get
the deprecation signal — the warning is loud where it shouldn't be and
silent where it should.

## Fix

Move the deprecation signal from module top into the relevant class
entry points (`HtmlDocument.__init__`, `ChunkedDocument.__init__`,
`SECHTMLParser.__init__`, and `Document.__post_init__` for the
`@dataclass`), via a small shared helper at
`edgar/files/_deprecation.py::warn_legacy_html_usage`.

The helper walks the call stack past the deprecation helper itself, the
three deprecated modules, and the `dataclasses` module (which hosts
synthesized `__init__` trampolines). If the first non-transparent frame
belongs to `edgar` or any `edgar.*` submodule, the call is internal and
the warning is suppressed. Any other caller — user code, notebooks,
tests, third-party libraries — receives the standard
`DeprecationWarning` pointing at its own call site.

This trades the "fires once at import" pattern for "fires per
user-driven instantiation," which is the standard Python pattern for
class-level deprecation and is what API consumers expect.

## Tests

`tests/test_internal_deprecation_silence.py`:

- `test_import_edgar_under_W_error` — subprocess: `python -W error
  -c "import edgar"` exits 0. Fails on `main`; passes with this change.
- `test_import_deprecated_submodules_under_W_error` — subprocess:
  importing each deprecated submodule directly under `-W error` is also
  silent.
- `test_user_instantiation_of_html_document_warns`,
  `test_user_instantiation_of_legacy_document_warns`,
  `test_user_instantiation_of_chunked_document_warns` — confirm that
  user-code instantiation still emits `DeprecationWarning`, so the
  user-facing deprecation notice is preserved.

All five pass with the fix; the import test fails on `main` (catches
the regression).

### DIFF
--- a/edgar/files/_deprecation.py
+++ b/edgar/files/_deprecation.py
@@ -1,0 +1,54 @@
+"""Frame-gated deprecation helper for legacy edgar.files.* modules.
+
+The legacy HTML stack (edgar.files.html, edgar.files.html_documents,
+edgar.files.htmltools) is deprecated in favor of edgar.documents, but
+edgartools' own code still instantiates these classes during normal
+operation. Naive module-top warnings.warn() calls fire on every
+internal import and pollute downstream test suites that run with
+``-W error`` — turning every ``import edgar`` into a failure.
+
+This helper suppresses the deprecation signal when the call site is
+edgartools-internal, and re-emits it for user code so the deprecation
+notice is preserved where it matters.
+"""
+
+import sys
+import warnings
+
+# Modules that are transparent to the caller check: the deprecated
+# modules themselves and (for dataclass-generated __init__ trampolines)
+# the standard library's dataclasses module. The dataclass machinery
+# emits an __init__ that runs in the *defining* module's namespace, so
+# we also skip the deprecated modules when walking up.
+_TRANSPARENT_MODULES = frozenset({
+    'edgar.files._deprecation',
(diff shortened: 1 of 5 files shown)


dgunning (approved):
Approved. Thanks for this PR

PR #845, opened 2026-06-08
fix: decode bytes HTML so SixK.text() handles bytes exhibits (#844)

### LINKED ISSUE #844: `SixK.text()` — `TypeError` When Parsing Certain 6-K Exhibits
## Issue Type

## Environment
- edgartools version: 5.35.1
- Python: 3.14
- OS: Windows 11

### Description

Calling `.text()` on a `SixK` object raises `TypeError: a bytes-like object is
required, not 'str'` for certain 6-K filings that contain exhibit attachments.
The error originates inside `edgar/files/html_documents.py` where a bytes/str
type mismatch occurs during HTML parsing.

### Reproduction

```python
import edgar

edgar.set_identity("user@example.com")

# Known problematic filing
filing = edgar.get_by_accession_number("0001104659-22-129316")  # FUTU 6-K
sixk   = filing.obj()

sixk.text()   # raises TypeError
```

### Expected Behavior

`sixk.text()` should return the text content of the 6-K filing as a string.

### Actual Behavior

```
TypeError: a bytes-like object is required, not 'str'
  File ".../edgar/company_reports/sixk.py", line 216, in _get_exhibit_content
    document = Document.parse(html_content)
  File ".../edgar/files/html_documents.py", line 460, in get_root
    if "<TEXT>" in html[:500]:
TypeError: a bytes-like object is required, not 'str'
```

### Workaround

Use `filing.text()` directly instead of `sixk.text()`:

```python
try:
    text = sixk.text()
except TypeError:
    text = filing.text()   # works correctly
```

---

gaoflow:
Fixes #844.

`SixK.text()` raises `TypeError: a bytes-like object is required, not 'str'` for 6-K filings whose exhibit content comes back as bytes.

`Attachment.download()` is typed `Optional[Union[str, bytes]]` and returns `bytes` for some exhibits. `SixK._get_exhibit_content` passes that straight into `Document.parse`, which reaches `HtmlDocument.get_root`:

```python
if "<TEXT>" in html[:500]:   # str in bytes -> TypeError
```

`BeautifulSoup` itself accepts bytes, so the only thing that breaks is the string-based `<TEXT>` SGML check above it. This decodes bytes once at the top of `get_root`, after which bytes and str inputs parse identically.

Reproducible offline (no SEC download needed):

```python
from edgar.files.html import Document
Document.parse(b"<html><body><p>hi</p></body></html>")   # was TypeError
```

**On the deprecation:** `edgar.files.html` / `edgar.files.html_documents` are marked deprecated for v6.0, but `SixK` (and the other `company_reports` objects) still route exhibit rendering through them on 5.x, so this is a live crash for current users. I put the fix at `get_root` since it's the shared crash site and covers every legacy-parser caller in one place — happy to move it to the `SixK._get_exhibit_content` boundary instead, or hold it, if you'd rather not touch the legacy parser.

Regression test in `tests/issues/regression/test_issue_844.py` (fast, no network): asserts `Document.parse` / `get_root` accept bytes, that bytes parses equivalently to str (both plain and `<DOCUMENT><TEXT>`-wrapped), and that non-UTF-8 bytes degrade gracefully. All four cases fail on `main` and pass here; the existing `tests/test_documents.py` offline suite is unchanged (63 pass — the 15 network/identity-gated failures are pre-existing and unrelated).

Thanks to @LuukBlom for the clear report and reproducer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`SixK.text()` no longer crashes on bytes exhibit content** — 6-K exhibits whose `Attachment.download()` returns `bytes` raised `TypeError: a bytes-like object is required, not 'str'` in the legacy HTML parser's `<TEXT>` check; the parser now decodes bytes first, so bytes and str inputs parse identically. ([#844](https://github.com/dgunning/edgartools/issues/844))
+
 ## [5.35.1] - 2026-06-04
 
 10-K section detection and agent TOC parsing receive two targeted fixes that close gaps introduced in 5.34.0.
--- a/edgar/files/html_documents.py
+++ b/edgar/files/html_documents.py
@@ -456,6 +456,10 @@ class HtmlDocument:
 
     @classmethod
     def get_root(cls, html: str) -> Tag:
+        # Attachment.download() returns str | bytes; decode bytes so the
+        # string-based <TEXT> checks below don't raise TypeError (GH #844)
+        if isinstance(html, (bytes, bytearray)):
+            html = html.decode("utf-8", errors="replace")
         # First check if the html is inside a <DOCUMENT><TEXT> block
         if "<TEXT>" in html[:500]:
             html = get_text_between_tags(html, 'TEXT')
--- a/tests/issues/regression/test_issue_844.py
+++ b/tests/issues/regression/test_issue_844.py
@@ -1,0 +1,42 @@
+"""
+Regression test for Issue #844: SixK.text() raises TypeError on bytes exhibit content.
+
+``Attachment.download()`` is typed ``str | bytes`` and returns bytes for some 6-K
+exhibits. That value flows into ``Document.parse`` -> ``HtmlDocument.get_root``,
+whose ``"<TEXT>" in html[:500]`` check raised
+``TypeError: a bytes-like object is required, not 'str'`` instead of parsing.
+
+The parser now decodes bytes before those string checks, so bytes and str inputs
+behave identically. No network access required.
+"""
+
+import pytest
+
+from edgar.files.html import Document
+from edgar.files.html_documents import HtmlDocument
+
+INNER_HTML = "<html><body><p>Exhibit 99.1 content</p></body></html>"
+SEC_WRAPPED = "<DOCUMENT>\n<TYPE>EX-99.1\n<TEXT>\n" + INNER_HTML + "\n</TEXT>\n</DOCUMENT>\n"
+
+
+class TestIssue844BytesExhibitContent:
+    """Parsing must accept bytes HTML, matching Attachment.download()'s str|bytes return."""
+
+    @pytest.mark.parametrize("html", [INNER_HTML, SEC_WRAPPED])
+    def test_document_parse_accepts_bytes(self, html):
+        """Document.parse must not raise TypeError on bytes input (GH #844)."""
+        doc_bytes = Document.parse(html.encode("utf-8"))
+        doc_str = Document.parse(html)
+        assert doc_bytes is not None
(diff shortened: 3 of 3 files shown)


warzoo:
Haha, thanks for the quick fix! Just to clarify, I'm actually just a regular user who ran into this bug, not the project maintainer, so I don't have the power to merge it myself. (And I think you meant to tag me, @warzoo, instead of LuukBlom! 😉)

The fix at get_root looks amazing and totally makes sense. Hopefully the project maintainers will review and merge it soon. Thanks again for your awesome work!

dgunning (commented):
## Review

Thanks for the careful write-up and the offline reproducer — the root-cause analysis is spot on, and putting the fix at `get_root()` (the shared crash site for every legacy-parser caller) is the right call.

### One requested change: encoding handling

`utf-8` + `errors="replace"` fixes the crash but silently corrupts non-UTF-8 exhibits, and older SEC filings are frequently cp1252/latin-1 (smart quotes, accented names). It's actually *worse* than handing the bytes to BeautifulSoup, which runs encoding detection. Verified locally:

```
input: cp1252 bytes for "café résumé — “quote”"
this PR (utf-8+replace): 'caf� r�sum� � �quote�'
BeautifulSoup raw bytes: 'café résumé — “quote”'   ← correct
```

So the change trades a *loud* crash for *silent* character corruption on exactly the legacy filings most likely to arrive as bytes, which cuts against the "data correctness is existential / no silent wrong data" principle.

Suggested fallback chain (cp1252 covers the SEC's dominant non-UTF-8 case; latin-1 never raises as a final guard):

```python
if isinstance(html, (bytes, bytearray)):
    for enc in ("utf-8", "cp1252"):
        try:
            html = html.decode(enc)
            break
        except UnicodeDecodeError:
            continue
    else:
        html = html.decode("latin-1")  # 1:1, never raises
```

This still fixes #844 and keeps UTF-8 clean, while decoding cp1252/latin-1 correctly instead of emitting `�`.

### Minor
- **Type hint:** `get_root(cls, html: str)` now accepts bytes — please annotate `Union[str, bytes]`.
- **Test masks the issue:** `test_non_utf8_bytes_do_not_crash` only asserts `is not None`, so it passes despite the corruption above. If you adopt the fallback decode, strengthen it to assert the characters survive (e.g. `"café" in` the parsed text) so it documents intent and guards the regression.
- **Scope (follow-up, not this PR):** only the legacy `edgar/files/` parser is fixed; the modern `edgar/documents/` parser isn't touched. Fine since `SixK` routes through the legacy path on 5.x — just flagging for a later check.

### Otherwise
Crash fix and placement are correct, regression test is well-structured (fast, offline, parametrized, asserts bytes≡str equivalence), CHANGELOG entry is clear. Performance/security are non-issues. Happy to approve once the fallback-decode (+ type hint + strengthened test) lands.

dgunning:
Thanks @gaoflow — this is a clean fix and I appreciate the offline reproducer and the note about the deprecated-parser routing. Decoding once at `get_root` is the right call: it's the shared crash site and covers every legacy-parser caller, not just `SixK`.

Merging as-is. I'm landing a small complementary commit on top that classifies `.xlsx`/`.zip` as binary in `binary_extensions` — those XBRL viewer artifacts (`Financial_Report.xlsx`, `<accession>-xbrl.zip`) were also slipping past the 6-K exhibit filter, so even without the crash they rendered empty/mojibake panels into `sixk.text()`. Your fix handles the crash; that handles the classification. Thanks again for the contribution! 🙏

binggao1230:
Thanks @dgunning — and good catch on the silent corruption; you're right that utf-8+replace is worse than handing bytes to BeautifulSoup for exactly these legacy filings. Opened #847 with the cp1252→latin-1 fallback, the `Union[str, bytes]` hint, and a strengthened test that actually asserts the characters survive.

And apologies @warzoo for tagging the wrong handle earlier — thanks for the report and the reproducer!

PR #847, opened 2026-06-09
fix: decode non-UTF-8 byte exhibits via cp1252/latin-1 fallback (#844 follow-up)

gaoflow:
Follow-up to #845, addressing @dgunning's review.

The merged fix decoded byte exhibits with `utf-8` + `errors="replace"`, which stops the crash but mojibakes cp1252/latin-1 filings (`café` → `caf�`) — trading a loud crash for silent corruption on exactly the legacy exhibits most likely to arrive as bytes. This applies the suggested fallback chain (utf-8 → cp1252 → latin-1; latin-1 never raises), so non-UTF-8 exhibits decode correctly.

- `get_root` is now annotated `Union[str, bytes]`.
- Strengthened the regression test: `test_non_utf8_bytes_preserve_characters` asserts `café`/`résumé` survive and no `�` appears (this fails on the old utf-8+replace), plus `test_undecodable_bytes_do_not_crash` guards the latin-1 final fallback with a byte invalid in both utf-8 and cp1252.

All offline, no network. The `edgar/documents/` modern parser is untouched (out of scope, as noted).

Credit correction: the reporter is @warzoo, not @LuukBlom — I tagged the wrong handle in #845, sorry about that, and thanks @warzoo for the report and reproducer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


dgunning (approved):
Approved. Thanks for this PR

PR #851, opened 2026-06-11
Add ISO 4217 currency column to XBRL facts DataFrame (#850)

### LINKED ISSUE #850: Feature Request: Normalize Currency Unit Identifiers in XBRL Facts to ISO 4217 Codes
## Feature Category

- [x] New API functionality
- [ ] Performance improvement
- [ ] Developer experience improvement
- [ ] Documentation enhancement
- [ ] Tool/utility addition

### Summary

`xbrl().facts.to_dataframe()` returns raw XBRL unit identifiers in the
`unit_ref` / `currency` field for non-USD companies, instead of standard
ISO 4217 currency codes. This makes currency-based filtering and display
unreliable for foreign companies.

### Current Behavior

```python
import edgar

edgar.set_identity("user@example.com")

filing = edgar.get_by_accession_number("0001104659-26-043451")  # FUTU 20-F
df = filing.xbrl().facts.to_dataframe()

# For a non-USD company, unit_ref looks like:
print(df[df["concept"].str.contains("Revenue")]["unit_ref"].iloc[0])
# Output: "UNIT_STANDARD_HKD_MNUSOXGRF0O9R60JINVDUQ"
# Expected: "HKD"
```

For comparison, a US company returns:

```python
filing = edgar.get_by_accession_number("0000320193-26-000007")  # AAPL 10-K
df = filing.xbrl().facts.to_dataframe()
print(df[df["concept"].str.contains("Revenue")]["unit_ref"].iloc[0])
# Output: "USD"  ← clean ISO 4217 code
```

### Affected Companies

Any non-USD company filing with SEC EDGAR. Confirmed affected:

| Company | Ticker | Currency | Raw unit_ref |
|---|---|---|---|
| Futu Holdings | FUTU | HKD | `UNIT_STANDARD_HKD_MNUSOXGRF0O9R60JINVDUQ` |
| (likely) Taiwan Semiconductor | TSM | TWD | (not yet tested) |
| (likely) any non-USD FPI | - | - | similar pattern |

### Root Cause

SEC XBRL allows companies to define custom unit identifiers. Foreign Private
Issuers (FPIs) sometimes use a UUID-suffixed format like
`UNIT_STANDARD_{CURRENCY}_{UUID}` instead of the plain ISO 4217 code.
edgartools currently passes these raw identifiers through without normalization.

### Proposed Solution

Add normalization logic to extract the ISO 4217 code from the raw unit
identifier:

```python
import re

def normalize_currency(raw: str) -> str:
    """Extract ISO 4217 currency code from XBRL unit identifier."""
    if not raw:
        return raw
    
    # Already a clean code (e.g., "USD", "HKD")
    if re.match(r'^[A-Z]{3}$', raw):
        return raw
    
    # Pattern: UNIT_STANDARD_{CURRENCY}_{UUID}
    match = re.search(r'UNIT_STANDARD_([A-Z]{3})_', raw)
    if match:
        return match.group(1)
    
    # Pattern: iso4217:USD
    match = re.search(r'iso4217:([A-Z]{3})', raw)
    if match:
        return match.group(1)
    
    return raw  # fallback: return as-is
```

### Suggested Implementation

Apply normalization in `FactsView.to_dataframe()` or in a dedicated
`get_currency()` helper method on the Filing object:

```python
# Option A: normalize in to_dataframe()
df["currency"] = df["unit_ref"].apply(normalize_currency)

# Option B: add a property to Filing
filing.reporting_currency  # returns "HKD", "TWD", etc.
```

### Impact

This affects any downstream application that:
- Filters facts by currency (e.g., `df[df["currency"] == "USD"]`)
- Displays currency to end users
- Stores currency in a database for later querying

Without normalization, developers must implement their own parsing logic,
leading to duplicated effort and potential inconsistencies.

### Workaround

```python
import re

def normalize_currency(raw: str) -> str:
    if not raw:
        return raw
    if re.match(r'^[A-Z]{3}$', raw):
        return raw
    match = re.search(r'UNIT_STANDARD_([A-Z]{3})_', raw)
    if match:
        return match.group(1)
    return raw

# Apply to dataframe
df["currency"] = df["unit_ref"].apply(normalize_currency)
```

### Environment

- edgartools version: 5.36.0
- Python: 3.14
- OS: Windows 11

---
*Feature requests are evaluated based on EdgarTools' core principles: Simple yet powerful, accurate financials, beginner-friendly, and joyful UX.*

gaoflow:
## Summary

Resolves #850. `xbrl().facts.to_dataframe()` exposed the raw XBRL `unit_ref` id for every fact. For non-USD filers that id is an opaque token such as `UNIT_STANDARD_HKD_MNUSOXGRF0O9R60JINVDUQ` instead of a usable currency, which made currency-based filtering and display unreliable for foreign companies (reported by @warzoo).

## What changed

The facts DataFrame now carries a **`currency`** column with each fact's ISO 4217 code (e.g. `USD`, `HKD`), resolved from the unit's `iso4217:` measure that the parser already records in `xbrl.units`. The opaque `unit_ref` is **preserved unchanged** — `currency` is purely additive:

```python
df = filing.xbrl().facts.to_dataframe()
df["unit_ref"]   # "UNIT_STANDARD_HKD_..."  (unchanged)
df["currency"]   # "HKD"                    (new)
```

- Per-share monetary units (e.g. `iso4217:USD` / `xbrli:shares`) report their numerator currency (`USD`).
- Non-monetary units (shares, pure, company-specific) resolve to `None` rather than a misleading value.

## Verification

`tests/issues/regression/test_issue_850.py`:
- **Issue repro** — a minimal instance using the opaque `UNIT_STANDARD_HKD_...` id resolves to `currency == "HKD"` while `unit_ref` stays opaque.
- **Ground truth** — the AAPL 10-K fixture resolves USD-denominated facts (and `usdPerShare`) to `"USD"`.
- **Silence check** — share-count facts have `currency` of `None`, not a fabricated code.

All three are red without the change. The existing facts/unit suites (`test_xbrl_unit_pointintime.py`, `test_xbrl_facts.py`, …) show no new failures (the remaining failures are pre-existing network-dependent tests). `ruff check`/`format` clean on the changed code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`currency` column on the XBRL facts DataFrame** — `xbrl().facts.to_dataframe()` now includes a `currency` column with each fact's ISO 4217 code (e.g. `USD`, `HKD`) resolved from its unit measure. Non-USD filers tag monetary facts with opaque unit ids such as `UNIT_STANDARD_HKD_MNUSOXGRF0O9R60JINVDUQ`, which were exposed verbatim in `unit_ref` and made currency-based filtering and display unreliable; the raw `unit_ref` is preserved, while `currency` gives a usable code. Per-share monetary units report their numerator currency, and non-monetary units (shares, pure, custom) resolve to `None` rather than a misleading value. ([#850](https://github.com/dgunning/edgartools/issues/850))
+
 ### Fixed
 
 - **TTM Q4 derivation no longer produces wrong/negative values for discrete-quarter reporters** — when a concept is reported as discrete quarters with no cumulative 9-month YTD fact (common for BDCs and investment companies), `TTMCalculator` derives Q4 as `FY - (Q1+Q2+Q3)`. It previously selected the three input quarters by their `fiscal_period` label, but the SEC tags comparative facts in re-filings with the *filing's* fiscal period, so the same calendar quarter could appear labeled Q1, Q2 and Q3 across successive 10-Qs — producing a wrong, often negative Q4 (e.g. GAIN `InvestmentCompanyDividendDistribution`: `57.2M - 3×28.8M = -29.2M`). Quarters are now selected by distinct calendar period (dedup by `period_end`, latest periodic filing wins), and derivation is skipped when a discrete Q4 is already reported. This affects `quarterize()`, TTM calculations, and quarterly statement views. ([#848](https://github.com/dgunning/edgartools/issues/848))
--- a/edgar/xbrl/facts.py
+++ b/edgar/xbrl/facts.py
@@ -43,6 +43,36 @@ def _deduplicate_facts(df: pd.DataFrame) -> pd.DataFrame:
     return df
 
 
+def _iso4217_code(measure: Optional[str]) -> Optional[str]:
+    """Return the ISO 4217 code of an ``iso4217:`` unit measure, else ``None``."""
+    if measure and measure.startswith('iso4217:'):
+        return measure[len('iso4217:'):]
+    return None
+
+
+def _unit_currency(unit_info: Optional[Dict[str, Any]]) -> Optional[str]:
+    """Resolve a parsed XBRL unit to its ISO 4217 currency code, or ``None``.
+
+    Currency facts use a simple ``iso4217:`` measure (e.g. ``iso4217:HKD`` ->
+    ``HKD``). Per-share monetary facts use a ``divide`` unit whose numerator is
+    the currency (e.g. ``iso4217:USD`` per ``xbrli:shares``), so the numerator
+    currency is returned. Non-monetary units (shares, pure, ...) return ``None``.
+    The opaque ``unit_ref`` id itself (e.g. ``UNIT_STANDARD_HKD_...``) is never
+    parsed -- only the resolved measure is used (see issue #850).
+    """
+    if not unit_info:
+        return None
+    unit_type = unit_info.get('type')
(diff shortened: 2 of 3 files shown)


dgunning (approved):
Approving — thanks for the clean, well-scoped contribution, @gaoflow.

This is a better solution than the one #850 proposed. Instead of regex-parsing the opaque `unit_ref` string (`UNIT_STANDARD_([A-Z]{3})_`), which is fragile and format-dependent, you resolve the currency from the already-parsed unit **measure** (`iso4217:HKD` → `HKD`) via `xbrl.units`. The opaque id is never parsed, only the structured measure — exactly the right source of truth.

What I verified locally:

- **Structure matches the parser** — `_unit_currency` handles both `{'type':'simple','measure':...}` and `{'type':'divide','numerator':[...]}`, which is exactly what `edgar/xbrl/parsers/instance.py:306-326` produces. The `simple`/`iso4217:` handling also mirrors existing code in `currency.py`.
- **Additive & non-breaking** — `unit_ref` is preserved verbatim; `currency` is purely new.
- **Tests pass** — all 3 in `tests/issues/regression/test_issue_850.py` (helper unit test, opaque-id repro → `HKD`, AAPL ground-truth + share-count `None` silence check). This covers our full Definition of Done: a ground-truth assertion, a verified repro, and a silence check.
- **No regressions** — `test_xbrl_facts.py` + `test_xbrl_unit_pointintime.py` (49 tests) stay green.
- CHANGELOG updated under `[Unreleased]`.

One optional, non-blocking follow-up: we can add a mention of the new `currency` column to the relevant skill YAML so AI agents discover it (our "solvability" principle) — happy to handle that separately so it doesn't hold up the merge.

Tracked internally as `edgartools-sv8d`. Will ship in the next release.

PR #852, opened 2026-06-11
[codex] Normalize XBRL currency unit identifiers

gyx09212214-prog:
## Summary
- Add `normalize_currency_unit()` to extract ISO 4217 codes from raw XBRL currency unit identifiers.
- Preserve raw `unit_ref` and add a normalized `currency` field to facts generated by `FactsView`.
- Cover `USD`, `iso4217:USD`, `UNIT_STANDARD_HKD_*`, and unknown unit formats.

## Why this matters
Foreign issuer filings can use raw unit identifiers like `UNIT_STANDARD_HKD_*`. A normalized `currency` column makes filtering, display, and database storage reliable without losing the original XBRL unit reference.

Refs #850.

## Validation
- `python -m py_compile edgar/xbrl/currency.py edgar/xbrl/facts.py tests/xbrl/test_currency_unit_normalization.py`
- Direct helper assertions passed.
- `git diff --check`

Note: full pytest was not run locally because this environment does not have `pytest` installed.

### DIFF
--- a/edgar/xbrl/currency.py
+++ b/edgar/xbrl/currency.py
@@ -24,6 +24,7 @@ Example usage:
     usd_revenue = converter.to_usd(dkk_revenue, 2024, rate_type='average')
 """
 
+import re
 from collections import Counter
 from dataclasses import dataclass, field
 from datetime import date, datetime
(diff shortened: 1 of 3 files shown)


dgunning:
Thanks for tackling this, @gyx09212214-prog — appreciate the contribution and the clean helper. We're going to land #850 via #851 instead, so I'm closing this one to avoid two implementations of the same feature.

The deciding factor is the resolution source. #851 reads the **parsed unit measure** (`xbrl.units[unit_ref]` → `iso4217:HKD`) rather than regex-matching the raw `unit_ref` string. That ends up mattering for three common cases this PR's string-based approach doesn't cover:

1. **US filers use a lowercase `usd` unit_ref** (this is what Apple and most domestic filers emit). `normalize_currency_unit("usd")` matches none of the branches (`[A-Z]{3}` is case-sensitive, no `iso4217:` / `UNIT_STANDARD_` prefix) and returns `"usd"` unchanged — so the most common case isn't normalized. Reading the measure yields `"USD"`.
2. **Per-share amounts** (`unit_ref="usdPerShare"`, a divide unit) return `"usdPerShare"` here; resolving the divide-unit numerator gives `"USD"`.
3. **Non-monetary facts** (shares, pure) return the raw token as a "currency"; the measure-based path returns `None`, which is what the issue asked for (no misleading value).

None of this is a knock on the work — the issue itself suggested the regex approach, and your tests are tidy. Please do keep contributing; happy to have you take a crack at other open items (e.g. the REIT subtype work in #854 is up for grabs).

PR #855, opened 2026-06-13
fix(formatting): roll format_currency_short up to the next unit at boundaries

gaoflow:
`format_currency_short()` picks the magnitude bucket from the raw value before rounding the mantissa to one decimal place. A value just under 1B therefore rounds to `1,000.0` within the millions bucket and renders as the nonsensical `$1,000.0M` instead of `$1.0B`:

```python
format_currency_short(999_950_000)  # '$1,000.0M'  -> should be '$1.0B'
format_currency_short(999_999_999)  # '$1,000.0M'  -> should be '$1.0B'
```

The fix promotes the value to the billions unit when the millions mantissa rounds up to 1,000. Values comfortably inside a bucket are unchanged (e.g. `999_000_000` stays `$999.0M`).

Added regression cases to `TestFormatCurrencyShort` covering the positive and negative just-under-1B values and a non-promoted high-millions value; the two rollover cases fail without the change.


dgunning (approved):
Approved. Thanks for this - the PR iw well done

dgunning (approved):
Approving — correct fix for a real rendering bug, thanks @gaoflow.

`format_currency_short()` selected the magnitude bucket from the raw value before rounding the mantissa, so a value just under 1B (>= ~999.95M) rendered as the nonsensical `$1,000.0M`. Promoting to `B` when `round(scaled, 1) >= 1000` is the right fix.

Verified:
- Float-precision at the boundary is exact: `round(999.95, 1) == 1000.0` and `round(999.949..., 1) == 999.9`, so the `>= 1000` check has no drift.
- All 15 `TestFormatCurrencyShort` cases pass (12 existing + 3 new), and the two rollover cases fail without the change.
- Minimal, consistent with the existing branch; matters in practice for the `to_context()` renderers in `form144.py` and `proxy/core.py`.

Non-blocking note: the analogous billions->trillions edge (e.g. `999_950_000_000` -> `$1,000.0B`) isn't handled, but there's no `T` unit to promote to, so that's pre-existing and out of scope.

Issue #457, opened 2025-10-11
Filing Access Failure: Crashing with ValueError on date parse


chuhanyi03:
## Issue Type

- [ ] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)  
- [x] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 4.18.0 (also in 4.17.1)
**Python Version:** 3.12.11
**Operating System:**  Windows 11

## Bug Description
**What you expected to happen:**
**In short, I hope these codes run normally without reporting errors.** 

**What actually happened:**
The library tried to parse this HTML and crashed with a ValueError because it found a date string it couldn't understand. It found a date string in my local language (Chinese) on the HTML page and crashed with a ValueError because the date format did not match what it expected. This error was misleading and made it very difficult to diagnose the real network problem. I tried in vain to change the network environment and force the locale to set an English time language environment, in an effort to ensure that the dates in the HTTP headers could be correctly parsed.

**Error message (if any):**
```
ValueError: time data '周五, 10 10月 2025 11:57:10 GMT' does not match format '%a, %d %b %Y %H:%M:%S GMT'
And it was displayed as follows once ValueError: 
time data 'Fri, 10 Oct 2025 10:54:12 GMT' does not match format '%a, %d %b %Y %H:%M:%S GMT'
And sometimes I find that the time displayed by the error message does not match the actual time. 
```

## Reproduction
**Company/Ticker:**  AAPL
**Time Period:** N/A. The issue occurs during the Company("AAPL") object initialization, before fetching filings for **any specific time period** .
**Relevant Forms:** all forms

**Minimal code to reproduce:**
```python

from edgar import Company
# method 1
apple = Company("AAPL")
filings = apple.get_filings(form="10-K")

#or method 2
form4_filings = get_filings(form="4")
form4 = form4_filings[0].obj()

#or method 3
filing = apple.latest("10-K")
tenk = filing.obj()

```

## Additional Context

<img width="1840" height="507" alt="Image" src="https://github.com/user-attachments/assets/1194c667-858a-4b7e-9569-069c2d74fd57" />

[...]

PR #858, opened 2026-06-14
fix: attach NullHandler to the edgar package logger (#856)

### LINKED ISSUE #856: Missing NullHandler in edgar package causes unexpected log output in application environments
## Issue Type

Environment

edgartools version: 5.36.0
Python: 3.14
OS: Windows 11

Issue 1: Missing NullHandler in edgar/__init__.py

Expected behavior (per [Python logging docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library))

It is strongly advised that you do not add any handlers other than NullHandler
to your library's loggers. [...] This is best practice that allows application
developers full control over logging output without having to silence spurious
output from libraries.

A well-behaved library should add a NullHandler at the package root:

python# edgar/__init__.py
import logging
logging.getLogger(__name__).addHandler(logging.NullHandler())

Actual behavior

edgar/__init__.py contains no logging configuration at all. All modules in the
package use log = logging.getLogger(__name__), which is correct, but without the
NullHandler at the package root, any log.warning() or log.error() call
propagates to the application's root logger.

If the application has a root logger with a StreamHandler (the default), all
edgartools warnings appear on stderr without the application having any way to
suppress them — short of manually calling:

pythonlogging.getLogger("edgar").setLevel(logging.CRITICAL)

Impact: MCP Server / stdio environments

This is particularly harmful in MCP (Model Context Protocol) server environments,
where stdout is a JSON-RPC communication channel. Any unexpected output on
stdout corrupts the protocol and causes the client to lose connection.

While the current warnings go to stderr (not stdout), the lack of NullHandler
means future print() calls or misconfigured handlers in any edgartools code
path could silently break MCP integrations. Several popular AI assistant tools
(including Claude Desktop) use edgartools via MCP.

---

gaoflow:
Fixes #856.

## Problem
The `edgar` package-root logger has no handler attached. When a consuming application has not configured logging itself, edgartools' `log.warning()` / `log.error()` calls have no handler anywhere in their ancestry, so Python falls back to the `logging.lastResort` handler, which writes to stderr at WARNING and above. The application has no clean way to suppress this short of `logging.getLogger("edgar").setLevel(logging.CRITICAL)`.

This is especially harmful in MCP / stdio environments, where stray stderr output corrupts the protocol stream.

## Fix
Attach a `logging.NullHandler` to the `edgar` logger, as recommended by the Python logging HOWTO ("Configuring Logging for a Library"). With a handler present in the ancestry, the `lastResort` fallback is not used, so edgartools stays silent until the application opts into logging. The library adds only a `NullHandler` and never a `StreamHandler`, leaving output control to the application.

The handler is attached after the package imports rather than at the top of `__init__.py`, so the existing module-level imports stay grouped at the top (avoids ruff E402).

## Verification
`tests/issues/regression/test_issue_856.py` asserts the `edgar` logger carries a `NullHandler` and no other handler type. Red before the change (the logger had no handlers), green after.

Implemented by me with AI assistance, under my direction and review.


dgunning (approved):
Approving — clean, correct, and exactly per the Python logging HOWTO. Thanks @gaoflow.

Verified locally:
- `logging.getLogger(__name__)` resolves to the `edgar` package-root logger (placement is after imports, before the #457 cache-clear block), and after import that logger carries **exactly one** handler — the `NullHandler`. Nothing else competes: `edgar/core.py` only calls `basicConfig` behind an env var and on the *root* logger, not `edgar`.
- Both regression tests pass (`test_issue_856.py`), including the stricter "only NullHandler attached" assertion.
- CHANGELOG entry is accurate (`logging.lastResort` → stderr fallback, MCP/stdio impact).

One note for tracking: this resolves **Issue 1** of #856. **Issue 2** — `sgml/sgml_header.py:827` using `log.warning()` for the known pre-2005 SGML header quirk, which should be `log.debug()` — is independently valid (apps that *do* configure logging still surface it) and will be handled as a separate follow-up. So I'll keep #856 open until that lands; this PR can merge on its own.

PR #860, opened 2026-06-15
fix: correct document search snippet offsets

gyx09212214-prog:
## Summary
- Adjust document search result offsets to point into the generated context snippet instead of the original full text.
- Preserve the existing `_get_context()` string-returning helper and add an internal helper that returns the context plus adjusted offsets.
- Add focused text and regex search regression tests for matches after a truncated prefix.

## Why this matters
`SearchResult.snippet` highlights `context[start_offset:end_offset]` when context is available. `_get_context()` already computed adjusted offsets for truncated contexts, but those adjusted values were discarded, so matches after the leading context window could highlight the wrong characters in snippets.

## Validation
- `python -m py_compile edgar/documents/search.py tests/test_textsearch_snippets.py`
- `git diff --no-index` checked that the patch only touches `edgar/documents/search.py` and adds `tests/test_textsearch_snippets.py`.

Full pytest was not run locally because git fetch/clone from this environment repeatedly timed out before a complete checkout could be created.

### DIFF
--- a/edgar/documents/search.py
+++ b/edgar/documents/search.py
@@ -181,12 +181,15 @@ class DocumentSearch:
                 flags = 0 if case_sensitive else re.IGNORECASE
 
                 for match in re.finditer(pattern, text, flags):
+                    context, start_offset, end_offset = self._get_context_with_offsets(
+                        text, match.start(), match.end()
+                    )
                     results.append(SearchResult(
                         node=node,
                         text=match.group(),
-                        start_offset=match.start(),
-                        end_offset=match.end(),
-                        context=self._get_context(text, match.start(), match.end())
+                        start_offset=start_offset,
+                        end_offset=end_offset,
+                        context=context
                     ))
             else:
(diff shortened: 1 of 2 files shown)


dgunning:
Thanks for this fix, @gyx09212214-prog — your diagnosis is exactly right: `_get_context()` computed context-relative offsets and then discarded them, so `SearchResult.snippet` highlighted the wrong characters for matches after a truncated prefix.

This is being superseded by #862, which lands the **same fix with the same approach** (a `_get_context_with_offsets()` helper feeding adjusted offsets into all three search paths, with `_get_context()` kept as a back-compat shim). I've folded your strongest contribution — the **end-to-end `search()` → `.snippet` regression tests** for text and regex modes — into #862, since they cover the real user-facing path better than testing the helper in isolation. Credit to you for that.

Closing this in favor of #862 to keep one PR; your work directly shaped the merged result. Much appreciated — hope to see more contributions.

PR #862, opened 2026-06-18
fix: use context-relative offsets in SearchResult snippet (fixes #860)

gaoflow:
_get_context computed adjusted offsets for truncated context but returned only the context string, discarding the adjustments. SearchResult.snippet then used the original (unadjusted) offsets against the truncated context, highlighting wrong characters for matches after a prefix context window.

Add _get_context_with_offsets that returns (context, adj_start, adj_end) and update text, regex, and whole-word search paths to use adjusted offsets.

Closes #860

### DIFF
--- a/edgar/documents/search.py
+++ b/edgar/documents/search.py
@@ -181,12 +181,15 @@ class DocumentSearch:
                 flags = 0 if case_sensitive else re.IGNORECASE
 
                 for match in re.finditer(pattern, text, flags):
+                    context, adj_start, adj_end = self._get_context_with_offsets(
+                        text, match.start(), match.end()
+                    )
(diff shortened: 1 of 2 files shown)


dgunning:
Pushed one commit (`b0e57a8`) adding end-to-end snippet assertions. The existing tests exercise `_get_context_with_offsets` directly (good isolation); these new ones drive the full `DocumentSearch.search()` path over a real `Document` and assert `SearchResult.snippet` highlights the correct characters in both text and regex modes — i.e. the actual user-facing surface. Ported from the equivalent tests in #860, which fixes the same bug with the same approach.

dgunning (approved):
Approved

Issue #863, opened 2026-06-19
Rule 10b5-1 detection ignores the official `<aff10b5One>` checkbox, silently returns `None` on real filings, and has both false positives and false negatives in its footnote matcher


warzoo:
## Issue Type

- [x] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Summary

`edgartools` detects Rule 10b5-1 trading plans purely by keyword-matching footnote
text (`detect_10b5_1_plan` in `edgar/ownership/core.py`). This approach has four
distinct problems that together make the result unreliable on real Form 4 filings:

1. **The authoritative source is ignored.** Since 2023-04-01 the SEC requires a
   document-level `<aff10b5One>` checkbox on Form 4/5 that states whether the trade
   was made under a Rule 10b5-1(c) plan. `edgartools` never reads this field, so for
   the very filings where a definitive answer exists, `has_10b5_1_plan` typically
   returns `None`.
2. **The footnote→transaction plumbing fails silently.** In the ownership-summary
   path, `footnote_ids` / `footnotes_text` are frequently left empty even when the
   filing clearly contains a 10b5-1 footnote, so `has_10b5_1_plan` returns `None`
   where it should return `True`.
3. **False positives:** the pattern list matches the *anti-fraud* Rule 10b-5 (a
   different rule), e.g. a footnote mentioning "anti-fraud Rule 10b-5" is reported
   as a 10b5-1 plan.
4. **False negatives:** plain substring matching misses common typographic variants
   ("10b5 1", "10b51", en/em-dash forms).

Each is reproduced below with real accession numbers.

## Environment

- **edgartools version:** 5.36.0
- **Python:** 3.14 (also reproduced on 3.12)
- **OS:** Windows 11

---

### Background

SEC Form 4/5 includes a document-level element `<aff10b5One>` (the "checkbox"
introduced by the Rule 10b5-1 amendments, mandatory for filings on/after
2023-04-01; some filers adopted it slightly earlier). Its value is the filer's
explicit, structured statement of whether the reported transactions were made
pursuant to a Rule 10b5-1(c) plan. Empirically the element contains either
[...]

PR #864, opened 2026-06-19
fix: use aff10b5One for 10b5-1 detection

gaoflow:
Addresses the structured-checkbox and matcher cases from #863.

Changes:

- Parse the Form 4/5 `aff10b5One` document-level checkbox into ownership objects.
- Have `TransactionSummary.has_10b5_1_plan` return that official value when present, before falling back to transaction footnotes.
- Tighten the footnote fallback matcher so bare anti-fraud Rule 10b-5 is not classified as a 10b5-1 trading plan, while common spacing/dash variants still match.
- Add a regression test for the checkbox precedence and matcher edge cases.
- This does not change the separate transaction footnote attribution path described in #863. 
Verification:`uv run --python 3.12 --with pytest pytest tests/issues/regression/test_issue_863_10b5_plan_detection.py tests/test_ownership_public_api.py -q`
- `uvx ruff check CHANGELOG.md edgar/ownership/core.py edgar/ownership/forms.py edgar/ownership/summary.py tests/issues/regression/test_issue_863_10b5_plan_detection.py`
-  `uv run --python 3.12 python -m compileall -q edgar/ownership tests/issues/regression/test_issue_863_10b5_plan_detection.py`
- `git diff --check`


binggao1230:
CI note: the package fast/slow jobs and CodeFactor are green. The only red job is test-network, which is also failing on current main (run 27782175061) and is failing in tests outside this change with VCR cassette / identity-related network errors. I do not see it tied to the 10b5 matcher change.

dgunning:
Thanks for this — clean fix, and I appreciate that you were explicit about scoping out the footnote-attribution path. I pushed one follow-up commit (`4f11748`) on top so the PR fully closes #863, since `maintainerCanModify` was enabled. Here's what it does and why.

### Why the extra commit

Your changes resolve Problems 1, 3 and 4. **Problem 2** (the `None`-when-a-10b5-1-footnote-exists case, e.g. AAPL `0000320193-22-000061`) was still open, because on pre-2023 filings with no `<aff10b5One>` element the per-transaction `footnotes_text` in the summary path is frequently empty — so `has_10b5_1_plan` returned `None` even though the filing's footnotes clearly describe the plan.

### What I added

- **`Footnotes.text`** (`models.py`) — combined text of all footnotes on the filing.
- **`forms.py`** — `get_ownership_summary()` now passes that full footnote text into `TransactionSummary` (`all_footnotes_text`).
- **`summary.py`** — `has_10b5_1_plan` falls back to scanning the full footnote set after the per-transaction check, before returning `None`. Precedence is unchanged in spirit:
  1. `<aff10b5One>` checkbox — authoritative when present (your logic)
  2. per-transaction footnote match **or** full-footnote-set scan
  3. `None` only when there's neither a checkbox nor any footnote evidence

  `None` now strictly means *"no checkbox and no footnote evidence"* rather than *"no plan."*

### One behavior change to flag

I removed the early `if not self.transactions: return None`. A filing can carry footnote evidence with an empty transaction list, and the early return would have discarded it; the new logic falls through to the footnote scan instead. Lower-risk than it looks, but it's the one spot where output can differ from your version, so wanted to call it out.

### Tests

- A **VCR-backed regression test** pinned to the real Problem 2 filing (AAPL insider Form 4 `0000320193-22-000061`): no checkbox, empty per-transaction attribution, but filing footnotes mention the plan → now `True` (was `None`). Cassette committed; replays offline.
- A **374Water fixture** end-to-end test for the checkbox path (the existing tests exercised the matcher and a synthetic summary, but not the `parse_xml → summary` wiring).
- Synthetic cases for the fallback (`True`/`False`/`None`) and checkbox-overrides-conflicting-footnotes.

Full file: **23 passed**, plus `test_ownership.py` / `test_ownership_public_api.py` green (no regressions).

### Not addressed (intentionally)

The *root cause* of Problem 2 — per-transaction footnote attribution arriving empty in the summary path — is worked around by the full-footnote scan, not repaired. Fixing that properly is a separate change in `summary_records.py` / the table parsing; happy to track it as a follow-up issue. Per-transaction granularity still lives on `TransactionActivity.is_10b5_1_plan` for callers who need it.

dgunning:
Opened #865 to track the root-cause attribution fix (per-transaction footnote IDs coming through empty in the summary path). This PR's full-footnote fallback stays as the safety net regardless.

dgunning:
Correction: the root-cause attribution follow-up is tracked internally in beads (`edgartools-t043`), not GitHub #865 (closed). This PR's full-footnote fallback stays as the safety net regardless.

dgunning (commented):
Approved

PR #896, opened 2026-07-11
fix: honor --runs in HTML parser benchmark

### LINKED ISSUE #895: HTML parser benchmark ignores --runs option
## Issue Type

- [x] Performance Issue (slow operations, memory problems)

## Environment
**EdgarTools Version:** 5.41.0
**Python Version:** 3.14.5
**Operating System:** Linux

## Bug Description
Passing `--runs N` to `tests/perf/benchmark_html_parser.py` should run each selected document exactly `N` times and record `"runs": N` in the output JSON.

**What actually happened:**
The CLI accepts `--runs`, but the value is not forwarded to `benchmark_corpus()`, which then calls `benchmark_document()` without it. Every document therefore uses the default five runs.

This makes quick benchmarks and higher-confidence repeated benchmarks misleading because the requested configuration is silently ignored.

**Error message (if any):**
```
No error message
```

## Reproduction
```bash
python tests/perf/benchmark_html_parser.py \
  --corpus data/html \
  --pattern OneTable.html \
  --runs 2 \
  --output /tmp/benchmark-results.json
```
The generated JSON reports:

"runs": 5

instead of:

"runs": 2

## Additional Context
The value must be propagated through both handoffs:

1. main() → ParserBenchmark.benchmark_corpus()
2. benchmark_corpus() → benchmark_document()

Regression tests should cover both handoffs.

---

JonArnfred:
Closes #895

### DIFF
--- a/tests/perf/benchmark_html_parser.py
+++ b/tests/perf/benchmark_html_parser.py
@@ -100,13 +100,14 @@ class ParserBenchmark:
             'runs': runs
         }
 
-    def benchmark_corpus(self, pattern: str = '*.html', max_files: Optional[int] = None) -> List[Dict]:
+    def benchmark_corpus(self, pattern: str = '*.html', max_files: Optional[int] = None, runs: int = 5) -> List[Dict]:
         """
         Benchmark all files in corpus.
 
         Args:
             pattern: Glob pattern for HTML files
             max_files: Maximum number of files to benchmark (None = all)
+            runs: Number of benchmark runs per file
 
         Returns:
             List of benchmark results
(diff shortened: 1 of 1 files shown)


dgunning (approved):
Clean, correct fix. Verified the `runs` value now threads through both handoffs (main → benchmark_corpus → benchmark_document), with the default of 5 preserved when `--runs` is omitted. Thanks for the well-documented issue and fix, @JonArnfred — a regression test covering both handoffs would be a nice follow-up per the issue notes, but not blocking for a test-harness script.

PR #897, opened 2026-07-15
fix: normalize MCP resource URIs

avielfedida:
## Summary

- accept the MCP SDK `AnyUrl` type in the resource handler
- normalize resource URIs to strings before matching known resources
- cover both documented resources with regression tests

## Problem

A standard MCP `ClientSession` can list both EdgarTools resources, but reading either listed URI fails with `Unknown resource`. The MCP SDK passes a Pydantic `AnyUrl` to registered resource handlers, and `AnyUrl` does not compare equal to the string literals currently used by the handler.

## Verification

- `pytest -q tests/test_mcp_server.py tests/test_mcp_intent_tools.py tests/test_ai_surface_verification.py` — 97 passed
- protocol-level stdio smoke test using `mcp.ClientSession` — both listed resources read successfully

### DIFF
--- a/edgar/ai/mcp/server.py
+++ b/edgar/ai/mcp/server.py
@@ -41,6 +41,7 @@ from mcp.server import NotificationOptions, Server
 from mcp.server.models import InitializationOptions
 from mcp.server.stdio import stdio_server
 from mcp.types import GetPromptResult, Prompt, TextContent
+from pydantic import AnyUrl
 
 # Set up logging
 logging.basicConfig(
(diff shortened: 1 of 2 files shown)


dgunning (approved):
Correct fix for a real protocol-compliance break — the MCP SDK passes `AnyUrl` to `read_resource`, so the string-literal comparison never matched and every documented resource read failed with `Unknown resource`. Normalizing via `str(uri)` is the right call, and the parametrized regression tests over both resources would have caught the original bug (ran them locally: 2 passed). Thanks @avielfedida — clean, well-verified first contribution.

PR #901, opened 2026-07-17
Perf/document parser hotpaths

### LINKED ISSUE #900: HTML parser performs unnecessary eager text rendering and expensive regex passes
## Performance Issue Details

Issue Type:

- [x] Slow operation (longer than expected)
- [ ] Memory usage too high
- [ ] Memory leak (increasing memory over time)
- [x] CPU usage too high
- [ ] Hangs/freezes during operation

## Environment

EdgarTools Version: 5.41.0
Python Version: 3.11.14
Operating System: Ubuntu 24.04.4 LTS
Available RAM: 96GB
CPU: Intel Core Ultra 9 185H

## Performance Metrics

Expected performance:

- Parsing should not render the complete document text unless text or statistics are requested.
- Preprocessing regexes should scale efficiently with filing size.
- Parsing representative local filings should be faster without changing extracted content.
- Memory: No specific memory issue was observed.

Actual performance:

Median of three alternating same-process runs:

Filing | Size | Time | 
Apple 10-K | 1.81 MiB | 785 ms 
Nvidia 10-K | 2.55 MiB | 1,110 ms
Oracle 10-K | 5.07 MiB | 1,892 ms
HealthPeak 424B2 | 0.90 MiB | 551 ms

## Reproduction Details

Company/Ticker: AAPL, NVDA, ORCL, HealthPeak
Dataset Size: Individual HTML filings between 0.90 and 5.07 MiB
Time Period: Checked-in HTML parser corpus

Code to reproduce:
```
from pathlib import Path
from statistics import median
from time import perf_counter

from edgar.documents import parse_html


def benchmark(path: str, runs: int = 5) -> float:
    html = Path(path).read_text()
    timings = []

    for _ in range(runs):
        start = perf_counter()
        parse_html(html)
        timings.append(perf_counter() - start)

    return median(timings)


for path in [
    "data/html/Apple.10-K.html",
    "data/html/Nvidia.10-K.html",
    "data/html/Oracle.10-K.html",
    "data/html/HealthPeak.424B2.html",
]:
    elapsed = benchmark(path)
    print(f"{path}: {elapsed * 1000:.1f} ms")
```
## Root Cause Analysis

Two main bottlenecks were identified.

### 1. Eager document statistics

DocumentPostprocessor._add_statistics() unconditionally calculates:

```"text_length": len(document.text())```

This renders and traverses the complete document, including tables, during every parse, even if the caller only needs sections, tables, search, or another structured operation.

The statistics are not serialized by DocumentMetadata.to_dict() and no internal consumers were found. A later call to document.text() also repeats part of this work.

Deferring statistics until document.metadata.statistics is first accessed substantially reduces default parse latency.

### 2. HTML preprocessing regexes

The current repeated-line-break pattern is:

``` r"(<br\s*/?>[\s\n]*){3,}" ```

It is expensive on large filings containing many isolated <br> tags. Across the 38-file HTML corpus, this pass took approximately 604 ms despite producing only two replacements.

## Suggested Fix

- Configure document statistics with a lazy, cached loader.
- Use a weak reference so the deferred loader does not retain discarded documents.
- Guard excessive-newline normalization with a cheap substring check.
- Replace the repeated break regex with the equivalent prefix-oriented expression.
- Replace the punctuation lookbehind with the verified equivalent lookahead form.

## System Resource Usage

- CPU usage: CPU-bound parser work, primarily regex substitution and eager text rendering.
- Memory usage: No abnormal growth observed.
- Disk I/O: Negligible; HTML is loaded before timing.
- Network usage: None; the benchmark uses local checked-in filings.

JonArnfred:
Closes #900

### Performance

Filing fixture | main | This PR | Improvement 
-- | -- | -- | -- 
Apple 10-K | 726.0 ms | 583.2 ms | 19.7% faster |
Nvidia 10-K | 1,113.9 ms | 870.1 ms | 21.9% faster |
Oracle 10-K | 1,871.1 ms | 1,534.2 ms | 18.0% faster |
HealthPeak 424B2 | 576.5 ms | 446.4 ms | 22.6% faster |

File loading and output validation were excluded from the parse timer.

Output equivalence was verified using the SHA-256 digest of the complete rendered text, table count, heading count, and document statistics.

### DIFF
--- a/edgar/documents/document.py
+++ b/edgar/documents/document.py
@@ -52,9 +52,10 @@ class DocumentMetadata:
         """Return document statistics, computing them on first access."""
         statistics = self.__dict__.get("_statistics")
         if statistics is None:
-            loader = self.__dict__.pop("_statistics_loader", None)
+            loader = self.__dict__.get("_statistics_loader")
             statistics = loader() if loader is not None else {}
             self.__dict__["_statistics"] = statistics
+            self.__dict__.pop("_statistics_loader", None)
         return statistics
 
     @statistics.setter
@@ -67,6 +68,12 @@ class DocumentMetadata:
         """Configure deferred statistics calculation for the owning document."""
         self.__dict__.pop("_statistics", None)
         self.__dict__["_statistics_loader"] = loader
+
+    def __getstate__(self) -> Dict[str, Any]:
+        """Materialize deferred statistics before serializing metadata."""
+        if "_statistics_loader" in self.__dict__:
+            _ = self.statistics
+        return self.__dict__.copy()
 
     def to_dict(self) -> Dict[str, Any]:
         """Convert metadata to dictionary."""
@@ -775,6 +782,12 @@ class Document:
     _text_cache: Optional[str] = field(default=None, init=False, repr=False)
     _config: Optional[Any] = field(default=None, init=False, repr=False)  # ParserConfig reference
     _section_extractor: Optional[Any] = field(default=None, init=False, repr=False)  # cached SECSectionExtractor
+
+    def __getstate__(self) -> Dict[str, Any]:
+        """Materialize lazy metadata before serializing a stable document state."""
+        if "_statistics_loader" in self.metadata.__dict__:
+            _ = self.metadata.statistics
+        return self.__dict__.copy()
 
     @property
     def sections(self) -> Sections:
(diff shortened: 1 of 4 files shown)


dgunning (approved):
Approved. Verified regex rewrites are behaviorally equivalent (0 mismatches across 400k differential-fuzz cases), the `\n{3,}` newline guard is idempotent-safe, lazy statistics have no non-test consumers and aren't serialized in `to_dict()`, and pickle round-trips correctly on a real 12MB filing via the `__getstate__` materialization. All 32 tests in test_html_parser.py pass. Nice, well-scoped perf work with thorough tests.

PR #902, opened 2026-07-17
fix(documents): preserve text following inline-XBRL facts

### LINKED ISSUE #898: filing.text() silently drops text following inline-XBRL facts (lxml .tail not read) — unit words like "billion" vanish from rendered text
### Summary

The new document parser used by `Filing.text()` (`edgar.documents.HTMLParser` → `rich_to_text`) silently
deletes the plain text that follows an inline-XBRL fact inside an inline container. The tagged value itself
survives; the text after the closing tag — typically the unit word and the rest of the sentence — disappears.

This turns, e.g.:

> `...these commitments were $95.2 billion, of which substantially all will be paid through fiscal year 2027.`

into:

> `...these commitments were $ 95.2`

Because the output still looks grammatical and complete, the loss is invisible downstream — but any consumer
reading dollar amounts from the rendered text now sees a bare number with its scale word deleted (95.2
"billion" reads as 95.2), and `ix:nonNumeric` facts can lose entire trailing clauses or sentences the same way.

### Minimal reproduction

```python
from edgar.documents import HTMLParser, ParserConfig
from edgar.richtools import rich_to_text

html = (
    '<html xmlns:ix="http://www.xbrl.org/2013/inlineXBRL"><body><div><span>'
    'commitments were $<ix:nonfraction name="us-gaap:OtherCommitment" contextref="c-1" scale="9">'
    '95.2</ix:nonfraction> billion, of which substantially all will be paid.'
    '</span></div></body></html>'
)
doc = HTMLParser(ParserConfig(form="10-K")).parse(html)
print(repr(rich_to_text(doc, width=500).strip()))
```

**Actual output** (identical on 5.12.0 and 5.42.0):

```
'commitments were $ 95.2'
```

**Expected output:**

```
'commitments were $95.2 billion, of which substantially all will be paid.'
```

### Real-world example

NVIDIA's FY2026 10-K (`https://www.sec.gov/Archives/edgar/data/1045810/000104581026000021/nvda-20260125.htm`).
The commitments note reads, in the source HTML:

```html
...these commitments were $<ix:nonFraction unitRef="usd" contextRef="c-177" decimals="-8"
   name="us-gaap:OtherCommitment" scale="9" id="f-894">95.2</ix:nonFraction>&#160;billion,
   of which substantially all will be paid through fiscal year 2027.
```

`filing.text()` renders this as `these commitments were $ 95.2` — the unit and the rest of the sentence are
gone. Measured across that one filing: 86 of 118 inline facts immediately followed by a scale word
("billion"/"million"/"thousand") lose it. Multi-fact sentences collapse into number runs — the cloud-commitments
sentence (`$27 billion, for which $7 billion, $6 billion, $5 billion, $5 billion, $2 billion, and $2 billion
will be paid...`) renders as `$ 27 7 6 5 5 2 2`. Even the cover page is affected: the registrant name is
marked up as `<ix:nonNumeric name="dei:EntityRegistrantName">NVIDIA CORP</ix:nonNumeric>ORATION`, and the
untagged tail `ORATION` is dropped, so the rendered text reads `NVIDIA CORP`.

The rate varies by filer (different HTML generators produce different inline-tag structures) — from ~0% on
some filings to ~78% of scale-word-adjacent facts on others — so spot-checks can easily miss it.

### Root cause

In lxml, text following an element's closing tag is stored on that element's `.tail`, not on the parent.
`_get_element_text()` in `edgar/documents/strategies/document_builder.py` collects `element.text` and each
child's `.text_content()`, but never reads the child's `.tail` — so for inline elements (which include
`ix:nonfraction` via `INLINE_ELEMENTS`), everything between the fact's closing tag and the next element is
dropped.

`lxml`'s own `text_content()` called on the *parent* handles this correctly (it stitches child tails in during
its recursive walk); the manual child-by-child walk in `_get_element_text` is what loses them.

Note the legacy renderer in the same package handles this correctly: `edgar.files.html_documents.html_to_text`
run on the same NVDA HTML loses 0 of the 118 scale words and outputs
`'commitments were $95.2 billion, of which substantially all will …'` — so a known-good reference
implementation already exists in the codebase.

### Environment

- Reproduced on edgartools **5.12.0** and **5.42.0** (latest at time of writing) — identical behavior.
- Python 3.13, macOS (Darwin 23.6).

### Two small adjacent observations (same files, possibly same fix pass)

- `INLINE_IXBRL_TAGS` in `edgar/files/html_documents.py` is defined but never referenced (dead code).
- In `document_builder.py`, `skip_header_detection_tags` contains `'ix:nonNumeric'` (mixed case) while tags
  are compared lowercased — that entry can never match.

Happy to provide more examples or test against a branch.

gottostartsomewhere:
### The bug

`filing.text()` (the `edgar.documents` parser) silently drops the text that follows an inline-XBRL fact inside an inline container. The tagged value survives, but the text after the closing tag, usually the unit word and the rest of the sentence, disappears. So a commitments note that reads "$95.2 billion, of which substantially all will be paid" comes out as "$ 95.2", and the reader sees a bare 95.2 with "billion" gone. On NVIDIA's FY2026 10-K the reporter measured 86 of 118 scale-word-adjacent facts losing their unit. Full credit to whoever wrote #898, that diagnosis is excellent and made this easy.

### Root cause and fix

In `_get_element_text` (`edgar/documents/strategies/document_builder.py`) the child loop collected each child's `text_content()` but never its `.tail`. In lxml, the text after an element's closing tag lives on that element's `.tail`, so for inline elements (which include `ix:nonfraction`) everything between the fact's close and the next element was dropped. The fix reads `child.tail` in that loop, including for skipped children, whose trailing text is still part of the document.

### Tests

Added `tests/issues/regression/test_issue_898_ixbrl_tail_text.py`, three network-free cases:

- text after an inline fact is preserved (the unit word and trailing clause come back)
- multiple facts in one sentence no longer collapse into a number run
- text after a skipped `ix:exclude` child is kept while the excluded content is dropped

I also ran the existing parser suite (`test_html_parser*.py`): 66 passed, 6 skipped, no regressions.

### One honest note on scope

This restores the dropped text. The exact inter-token spacing ("$ 95.2 billion" vs "$95.2 billion") is a separate, pre-existing thing: `preprocessor._normalize_whitespace` strips whitespace adjacent to tags (`spaces_after_tags`), which on the unmodified parser also merges normal inline words (e.g. "the quick brown fox" already renders as "The quick brownfox"). That is a broader string-level change with a wider blast radius, so I kept it out of this PR to stay focused on the data loss. Happy to tackle it separately if you want it.

The issue also flagged two small adjacent items (dead `INLINE_IXBRL_TAGS` in `html_documents.py`, and a mixed-case `ix:nonNumeric` in `skip_header_detection_tags` that can never match). I confirmed both. Glad to fold them into this PR or a follow-up, whichever you prefer.

Closes #898

### DIFF
--- a/edgar/documents/strategies/document_builder.py
+++ b/edgar/documents/strategies/document_builder.py
@@ -558,36 +558,46 @@ class DocumentBuilder:
 
     def _get_element_text(self, element: HtmlElement) -> str:
         """Get text content from element."""
+        inline = element.tag.lower() in self.INLINE_ELEMENTS
         text_parts = []
 
         # Get element's direct text
         if element.text:
             # For inline elements, preserve leading/trailing whitespace
-            if element.tag.lower() in self.INLINE_ELEMENTS:
+            if inline:
                 text_parts.append(element.text)
             else:
                 text_parts.append(element.text.strip())
 
         # For simple elements, get all text content
-        if element.tag.lower() in self.INLINE_ELEMENTS or \
-           element.tag.lower() in ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']:
+        if inline or element.tag.lower() in ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']:
             # Get all text including from child elements
             for child in element:
                 if child.tag.lower() not in self.SKIP_ELEMENTS:
                     child_text = child.text_content()
                     if child_text:
                         # For inline elements, preserve whitespace in child content too
-                        if element.tag.lower() in self.INLINE_ELEMENTS:
+                        if inline:
                             text_parts.append(child_text)
                         else:
                             text_parts.append(child_text.strip())
+                # Text following a child's closing tag lives on child.tail (lxml),
+                # not on the parent. Without capturing it, the unit word and the
+                # rest of the sentence after an inline fact (e.g. ix:nonfraction)
+                # are silently dropped. Read it even for skipped children, since
+                # their trailing text is still part of the document. (gh-898)
+                if child.tail:
+                    if inline:
+                        text_parts.append(child.tail)
+                    elif child.tail.strip():
+                        text_parts.append(child.tail.strip())
 
-        # For inline elements with preserved whitespace, concatenate directly
-        # For others, join with spaces
-        if element.tag.lower() in self.INLINE_ELEMENTS and len(text_parts) == 1:
-            return text_parts[0] if text_parts else ''
-        else:
-            return ' '.join(text_parts)
+        # A single preserved inline part is returned as-is; otherwise space-join
+        # so word boundaries survive (the preprocessor strips whitespace adjacent
+        # to tags, and this re-separates the parts).
+        if inline and len(text_parts) == 1:
+            return text_parts[0]
+        return ' '.join(text_parts)
 
     def _is_text_only_container(self, element: HtmlElement) -> bool:
         """Check if element contains only text and inline elements."""
--- a/tests/issues/regression/test_issue_898_ixbrl_tail_text.py
+++ b/tests/issues/regression/test_issue_898_ixbrl_tail_text.py
@@ -1,0 +1,74 @@
+"""
+Regression test for Issue #898: text after inline-XBRL facts is silently dropped.
+
+GitHub Issue: https://github.com/dgunning/edgartools/issues/898
+
+Bug (FIXED): the document parser's ``_get_element_text()`` collected each child's
+``text_content()`` but never its ``.tail``. For inline containers this dropped the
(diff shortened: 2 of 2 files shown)


dgunning (approved):
Approved. Traced the call graph to confirm no double-counting: `_get_element_text` only feeds TextNode/HeadingNode paths, and `_should_process_children` returns False for those, so captured child `.tail`s are collected exactly once. Differential on real fixtures confirms it: Nvidia 'billion' 53→84 and Apple 27→82 (recovered scale words), while Oracle and JPM come out byte-identical — surgical recovery, no duplication. 27 tests pass. Well-scoped fix with an exemplary write-up.

Issue #905, opened 2026-07-24
TenQ item-boundary detection scatters MD&A into a phantom "Part I, Item 6" (Freddie Mac 10-Q)


sf1tzp:
## Issue Type

- [x] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.43.0
**Python Version:** 3.13.2
**Operating System:** Ubuntu 24.04.4 LTS

## Bug Description

**What you expected to happen:**

- No phantom `part_i_item_5` / `part_i_item_6` for a 10-Q (Part I has Items 1–4 only).
- `Part I, Item 2` returns the full MD&A (the ~165k currently captured under the phantom item).
- `Part I, Item 4` returns the Controls & Procedures section, not an MD&A table.

**What actually happened:**

For a Freddie Mac (FMCC) 10-Q, `TenQ` section/item detection mis-aligns the Part I item boundaries. The bulk of MD&A (~165k chars) is assigned to a **phantom `part_i_item_6`** — a 10-Q Part I only has Items 1–4 — while the real items return tiny, mislabeled scraps:

- `Part I, Item 2` (MD&A) returns a 1,264-char fragment (a single net-interest-income table).
- `Part I, Item 4` (Controls & Procedures) returns a *non-interest-income* MD&A table — wrong section content entirely.

The likely trigger is the filing's repeating page headers ("Management's Discussion and Analysis" appears 36 times as a running header), which the boundary detector appears to treat as section starts.

**Error message (if any):**
```
None — the content is silently mis-attributed; the phantom items carry no warnings.
```

## Reproduction
**Company/Ticker:** FEDERAL HOME LOAN MORTGAGE CORP (FMCC), CIK 0001026214
**Time Period:** Q1 2026 (period 2026-03-31, filed 2026-04-30)
**Relevant Forms:** 10-Q, accession 0001026214-26-000027

**Minimal code to reproduce:**
```python
import os
os.environ.setdefault("EDGAR_IDENTITY", "you your.email@example.com")
from edgar import find

f = find("0001026214-26-000027")
obj = f.obj()                       # TenQ
[...]

Issue #891, opened 2026-07-09
Item 7/Item 1 undetected when the bare "Item N." fragment and its title are in separate nodes; heading strategy also drops sections by leaving .item/.part unset


abatkins:
## Summary

On a small-cap 10-K whose section headers put the bare **`Item 7.`** fragment and the
**`Management's Discussion and Analysis…`** title in **separate HTML nodes**, `HybridSectionDetector`
detects **no MD&A (Item 7) section at all** (Item 1/Business fails the same way). Two independent
gaps combine to produce this, and a third makes the *heading* strategy silently lossy even when it
does fire:

1. **Heading strategy needs the title in the same string.** `HeaderDetector`'s primary item regex is
   `^(Item|ITEM)\s+(\d+[A-Z]?)[.\s]+(.+)$` — the trailing `(.+)$` requires non-empty title text in the
   same node. A bare `"Item 7."` node fails it, so `header_info.is_item` is never set True, and
   `HeadingSectionDetector` skips it (`if not header_info.is_item: continue`).

2. **Pattern strategy's zero-candidate case never reaches its own HTML fallback.** The `mda` patterns
   (`^(Item|ITEM)\s+7\.?\s*Management.*Discussion`, `^Management.*Discussion.*Analysis`, `^MD&A`) don't
   match the bare `"Item 7."`, so `mda` collects **zero** candidates. `SectionExtractor._match_sections`
   wraps the recovery in `if candidates:`, so `_find_actual_section_after_toc` (which *does* search the
   raw HTML for a bare `ITEM\s+7\.`) is never called when a section has no candidates at all — exactly
   the case that needs it.

3. **Heading-detected sections lose `.item`/`.part`.** `HybridSectionDetector._extract_section_from_heading`
   names the section `item_{N}` (no part) and constructs `Section(...)` **without** passing `part=`/`item=`,
   so both default to `None` even though `header_info.item_number` was known. Any consumer keying on
   `section.item` (e.g. building a per-Item map) silently drops these sections. (Contrast the pattern and
   TOC extractors, which set `part`/`item` via `Section.parse_section_name`.)

## Reproduction

- Filing: Nathan's Famous, Inc. (NATH), CIK **0000069733**, 10-K accession **0001437749-26-019923**.
- edgartools **5.37.0** (root cause re-verified against current `main`, past 5.41.0).

```python
from edgar import Company
tenk = Company("0000069733").get_filings(form="10-K").latest().obj()
[...]

Issue #880, opened 2026-06-28
`TenK[item]` merges adjacent Part III items; the absorbed item returns empty


warzoo:
## Issue Type

- [x] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Summary

`TenK[item]` returns text for an item by locating its header and reading to the
next item boundary. For some filings the parser fails to detect the next Part III
item header, so one item absorbs the following one and the following item returns
empty.

Example: Tesla's FY2022 10-K, accession `0000950170-23-001409` (filed
**2023-01-31**). `TenK["Item 10"]` returns 687 chars that include the
`ITEM 11. EXECUTIVE COMPENSATION` header and its body (at offset 515), while
`TenK["Item 11"]` returns `None`/empty. Note `TenK.items` does not list any
Part III item at all for this filing (`['Item 1', 'Item 1A', 'Item 2', 'Item 3',
'Item 7', 'Item 8', 'Item 15']`), yet `TenK["Item 10"]` resolves via the section
parser.

Same family as the other boundary issues (`bug_s1_section_boundaries.md`,
`bug_8k_signature_block_in_last_item.md`): the content is present, the section
boundary is wrong. Markup-dependent — a 10-K/A Part III amendment with real bodies
splits cleanly (`0001104659-25-042659`: Item 11 = 75,881 chars in its own result).

## Environment

- **edgartools version:** 5.40.0
- **Python:** 3.14
- **OS:** Windows 11

## Reproduction

```python
import edgar
edgar.set_identity("you@example.com")

# Tesla FY2022 10-K (filed 2023-01-31). Part III is incorporated by reference.
o = edgar.get_by_accession_number("0000950170-23-001409").obj()

r10, r11 = o["Item 10"], o["Item 11"]
t10 = str(r10).strip() if r10 else ""
t11 = str(r11).strip() if r11 else ""
print("Item 10 len            :", len(t10))
print("Item 11 len            :", len(t11))
print("Item 10 contains ITEM 11:", "ITEM 11" in t10.upper())
print("  'ITEM 11' offset       :", t10.upper().find("ITEM 11"), "of", len(t10))
```

### Actual

```
[...]

Issue #836, opened 2026-06-03
10-K section detector creates spurious `part_iv_item_1` / `part_iv_item_1a` sections (UNH 2026 10-K); `sections.get_item("1")` returns 220K of Risk Factors instead of Item 1 Business


FlorinAndrei:
## Issue Type

- [x] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.35.0
**Python Version:** 3.13.13
**Operating System:** Ubuntu 24.04.4 LTS (aarch64)

## Bug Description

This is closely related to (and a follow-on from) #821. The `.business` property itself is now **correct** for this filing — it returns the real Part I Item 1. The remaining problem is in the underlying section index that `document.sections` exposes.

**What you expected to happen:**

For a 10-K, the section collection should contain Item 1 / Item 1A only under Part I. A 10-K's Part IV contains Items 15 and 16 only — there is no "Part IV Item 1" or "Part IV Item 1A". Accordingly:
- `document.sections` should not contain `part_iv_item_1` / `part_iv_item_1a` keys.
- `sections.get_item("1")` (documented as "Returns first Item 1 (any part)") should return the Item 1 Business section.

**What actually happened:**

For UnitedHealth's 2026 10-K, the TOC section detector creates two spurious sections, each 220,642 chars, both starting at "ITEM 1A. RISK FACTORS" and running to near the end of the document (~4.7× the size of the real Item 1):

| key | part | item | length | starts with |
|-----|------|------|--------|-------------|
| `part_i_item_1`   | I  | 1  | 46,852  | `ITEM 1.  BUSINESS / OUR BUSINESSES / Overview …` ✅ correct |
| `part_iv_item_1`  | **IV** | 1  | 220,642 | `ITEM 1A.  RISK FACTORS / CAUTIONARY STATEMENTS …` ❌ |
| `part_iv_item_1a` | **IV** | 1A | 220,642 | `ITEM 1A.  RISK FACTORS / CAUTIONARY STATEMENTS …` ❌ (duplicate of the above) |

[...]

PR #906, opened 2026-07-24
fix(sections): repair TOC-subset overflow (Coeur 10-K) and flag schema-invalid 10-Q items (FMCC) — GH #904, #905

### LINKED ISSUE #904: TenK Item 7 overflows to end-of-document when the TOC parse only anchors a subset of items (Coeur Mining 10-K)
## Issue Type

## Environment
**EdgarTools Version:** 5.43.0
**Python Version:** 3.13.2
**Operating System:** Ubuntu 24.04.4 LTS

## Bug Description

`part_ii_item_7`'s end boundary should not default to end-of-document when the TOC parse fails to anchor subsequent items — the body headers for Items 7A/8/9A are present and line-anchored (`"Item 7A.\xa0\xa0...Quantitative and Qualitative Disclosures..."`), so body-header recovery has clean boundaries to use. Alternatively, when a pattern-detected named section (`mda`) exists and the TOC-detected equivalent is anomalously larger, the size anomaly should at least surface in `Section.warnings`.

**What actually happened:**

For Coeur Mining's FY2025 10-K, `TenK['Item 7']` (and the `management_discussion` property) returns **257,440 chars running from the Item 7 header to the director signatures at the end of the filing**, with Items 7A, 8, 9A, 10 and 11 embedded inside. The real MD&A is ~124k chars.

The cause is visible in the section map: the TOC-based detection pass (confidence 0.95) only anchored a subset of items — `part_ii_item_7` is the **last** TOC-detected section, with no entries for 7A/8/9/9A or Parts III/IV — so its end boundary defaults to end-of-document. The same happens to `part_i_item_1b` (317k chars for what should be a one-line "Unresolved Staff Comments" stub).

Notably, the **pattern-based pass got it right**: `document.sections['mda']` is a correctly bounded 123,925-char section (starts at the MD&A title, contains no later item headers, ends where Item 7A begins). But the item-lookup layer prefers the higher-confidence TOC result, so the good extraction loses the tiebreak. Neither the body-header item recovery nor `Section.warnings` anomaly flagging fires — `warnings` is `[]` on every section despite the 2x/470x size anomalies.

**Error message (if any):**
```
None — the over-capture is silent (confidence 0.95, warnings == [] on every section).
```

## Reproduction
**Company/Ticker:** Coeur Mining, Inc. (CDE), CIK 0000215466
**Time Period:** FY2025 (period 2025-12-31, filed 2026-02-18)
**Relevant Forms:** 10-K, accession 0000215466-26-000004

```python
import os
os.environ.setdefault("EDGAR_IDENTITY", "you your.email@example.com")
from edgar import find

f = find("0000215466-26-000004")
tenk = f.obj()
print("full text:", len(f.text()))        # 563,002

# Item lookup returns the overflowed TOC section:
mda = tenk["Item 7"]
print(len(mda))                            # 257,440
print(mda[-200:])                          # ...director signature block (end of 10-K)

# Later item headers are embedded inside "Item 7":
import re
for item in ["Item 7A", "Item 8", "Item 9A"]:
    m = re.search(rf"(?m)^\s*{item}\s*\.", mda)
    print(item, "->", m.start() if m else None)

# The pattern-detected named section is correct:
print(len(dict(tenk.document.sections.items())["mda"].text()))   # 123,925

# Section map shows the incomplete TOC parse:
for name, s in tenk.document.sections.items():
    print(name, len(s.text() or ""), getattr(s, "confidence", None),
          getattr(s, "detection_method", None), getattr(s, "warnings", None))
```

Observed:

```
full text: 563002

tenk["Item 7"]: 257440 chars
  embedded headers: Item 7A @ 91688, Item 8 @ 99357, Item 9A @ 232505,
                    Item 10 @ 237351, Item 11 @ 238015
  tail: "...J. Kenneth Thompson  Director  February 18, 2026"   (end of filing)

section map (name: chars, confidence, method):
  part_i_item_1:    25622  0.95 toc
  part_i_item_1a:  100773  0.95 toc
  part_i_item_1b:  317604  0.95 toc      <- overflow (should be a stub)
  part_i_item_1c:    4474  0.95 toc
  part_i_item_2:    53892  0.95 toc
  part_i_item_4:      252  0.95 toc
  part_ii_item_5:    1529  0.95 toc
  part_ii_item_7:  257440  0.95 toc      <- overflow; last TOC anchor, runs to EOF
  business:        126529  0.63 pattern
  risk_factors:    100951  0.63 pattern
  mda:             123925  0.70 pattern   <- correct MD&A
  market_risk:       7650  0.70 pattern
  financial_statements: 202560 0.63 pattern
  controls_procedures:  192365 0.63 pattern
  ...
  (warnings == [] on every section)
```

### Root cause (two gaps that combine)

**1. `_is_bold_header` misses split-span headers** — `TOCAnalyzer._is_bold_header` (`edgar/documents/utils/toc_analyzer.py`) only checks the element's *own* inline style (h1–h6 tag, or `font-weight:bold|>=600` in `el.get('style')`). Coeur's body item headers are a `<div style="margin-top:14pt">` whose weight lives on child spans — and the spans are split so that only the *title* span is bold:

```html
<div style="margin-top:14pt">
  <span style="...font-weight:400...">Item 7A.</span>
  <span style="...font-style:italic;font-weight:400...">&#160;&#160;...</span>
  <span style="...font-weight:700...;text-decoration:underline">Quantitative and Qualitative Disclosures About Market Risk</span>
</div>
```

So the body-header scan matches **zero** headers on this filing — calling `_analyze_body_item_headers` directly returns `{}`. The bold-child recognition added in 5.40.1 for the section-pattern strategies (GH #880) was not applied to this scanner.

**2. Body-header recovery is all-or-nothing behind a floor gate** — in `TOCAnalyzer.analyze_toc_structure`:

```python
if self._canonical_item_count(result) < self._expected_item_floor():   # floor = 8 for 10-K
    body = self._analyze_body_item_headers(html_content, tree=tree)
    if self._canonical_item_count(body) > self._canonical_item_count(result):
        return body                                                    # wholesale replace
```

Coeur's TOC parse anchors **exactly 8** canonical items (1, 1A, 1B, 1C, 2, 4, 5, 7) → `8 < 8` is false → recovery never runs even once gap 1 is fixed. And because it *replaces* rather than merges, a partially-good TOC parse can't be repaired by a partially-good body scan. A union-merge (run the body scan when the map is gappy, fill in only the items the TOC missed, TOC wins conflicts) would fix this class generally: each Coeur body header has an anchor div with an id a few siblings before it (e.g. `i96e67f5c3ce24f5882bca23c01e2ca6f_115`), so the existing `{section_key: anchor_id}` contract works once the header matches.

Direct harness showing the gate:

```python
from edgar.documents.utils.toc_analyzer import TOCAnalyzer
a = TOCAnalyzer(form="10-K")
toc = a.analyze_toc_structure(html)                # 8 items, ends at part_ii_item_7
body = a._analyze_body_item_headers(html)          # {} (should find 7A/8/9A/10/11)
print(a._canonical_item_count(toc), a._expected_item_floor())   # 8, 8
```

### Impact

Downstream consumers reading `tenk["Item 7"]` get 2x the real MD&A plus the entire back half of the 10-K (financial statements, controls, Part III stubs, signatures), which silently duplicates content across sections and inflates token counts for NLP/RAG pipelines. Because the TOC pass reports confidence 0.95 and no warnings, there is no signal to detect the over-capture without scanning for embedded item headers.

### Related prior art (all distinct failure modes)

- #891 (fixed 5.42.0) — split-node `Item 7.` headers in the pattern/heading strategies; the same split-styling family, but the TOC body scanner still has the single-element assumption.
- #880 (fixed 5.40.1) — bold-child heading recognition for Part III boundaries ("Strategy 3b"); not applied to `_is_bold_header` in the TOC analyzer.
- #873, #836 — other TOC/section boundary repairs; different mechanisms.

Standard filers (e.g. AAPL FY2025 10-K, `0000320193-25-000079`) parse correctly on 5.43.0 — all items bounded, including Part III stubs. Happy to provide additional problem filings if useful, and **happy to submit a PR** for the two-part fix (bold-child recognition in `_is_bold_header` + union-merge recovery) with regression tests against this filing plus AAPL/Citigroup controls.

---

sf1tzp:
Fixes #904. Addresses #905 (flags the phantom items; the underlying running-header mis-anchoring is a separate problem — details below).

## What this fixes

**#904 — Coeur Mining FY2025 10-K (`0000215466-26-000004`):** `TenK['Item 7']` returned 257,440 chars running from the Item 7 header to the director signatures at end-of-document, with Items 7A/8/9A/10/11 embedded; Item 1B returned 317,604 chars for a one-line "None." stub. Three gaps combined, each fixed in its own commit:

1. **Split-span bold headers** (`toc_analyzer.py`) — `_is_bold_header` only checked the element's own inline style. Coeur's body item headers are unstyled divs whose font weight lives on child spans, split so only the *title* span is bold ("Item 7A." at weight 400, the title at 700), so `_analyze_body_item_headers` matched zero headers. The check now also accepts an element whose bold child spans carry at least half of its text (same spirit as the 5.40.1 bold-child heading strategy from #880, applied to this scanner).
2. **Union-merge body-header recovery** (`toc_analyzer.py`) — recovery ran only below the 10-K item floor (8) and replaced the map wholesale. Coeur anchored exactly **8** canonical items, so recovery never fired — and Item 7, the last anchor, extended to EOF. When core items every 10-K carries (1, 1A, 7, 7A, 8, 9A) are missing, the body scan now runs and fills in **only the items the TOC missed**; the TOC stays the higher-trust source on conflicts. The wholesale-replacement floor path (GS/Citi class) is untouched.
3. **Shared-anchor end boundaries** (`toc_section_extractor.py`) — Coeur's one-line Item 1B stub and Item 1C target the *same* anchor div, so Item 1B's end anchor equalled its start anchor — an empty span the slicer treats as unbounded (ran to EOF). End boundaries now come from the first following section whose anchor differs.

**#905 — Freddie Mac Q1 2026 10-Q (`0001026214-26-000027`):** MD&A (~165K chars) scattered into a phantom `part_i_item_6` while the real Part I items returned mislabeled scraps — at 0.95 confidence with no warnings (the size bands catch `part_i_item_6` on 5.43.0, but `part_i_item_5` and the schema-impossibility itself were silent). Per the flag-and-return principle (Verification Constitution #2 / D4), two guardrails are appended to `HybridSectionDetector._validate_pipeline`, after the confidence filter like the size guardrail so a downgrade cannot silently drop the flagged section:

4. **Per-part item validity** — the 10-Q `FormSchema` now declares `part_item_ranges` (Part I: Items 1–4, Part II: 1–6). A detected key outside its Part's range cannot exist on the form; it now carries a warning and `ANOMALOUS_CONFIDENCE`. Content is kept — dropping it would lose the only copy of the MD&A text, and a schema-invalid key is precisely the signal that the whole Part's boundaries are suspect. This is the 10-Q counterpart of the #836 directional part validation (10-K validity is already implied by `item_part_ranges` at key construction).
5. **Successor-header guardrail** — a large TOC item section whose text embeds a line-anchored header of a later item *absent from the map* has an end boundary that ran past it — the over-capture class size bands can't see (Item 7's band is generous; Item 1B has no band at all). Scoped to unique-item forms (10-K), gated on a missing later item *number* so fully-anchored filings pay **zero** extraction cost, with a density guard so embedded mini-TOC/index pages don't false-positive.

To be explicit about #905's scope: this PR makes the FMCC failure *detectably* wrong (warnings + reduced confidence on the phantom keys) rather than silently wrong. The running-header mis-anchoring that scatters the content in the first place is not attempted here.

## Validation

| Filing | Before | After |
|---|---|---|
| Coeur 10-K `Item 7` | 257,440 chars, ends at signatures | **91,682**, bounded at Item 7A, no embedded headers |
| Coeur 10-K `Item 1B` | 317,604 chars | **4,474** (stub + adjacent 1C, shared anchor) |
| Coeur 10-K 7A/8/9A/9B/9C/10/14 | absent from TOC map | recovered as `toc` sections |
| FMCC 10-Q `part_i_item_5`/`6` | 0.95, silent | flagged, `confidence=0.5`, "Part I of a 10-Q has no Item N" |
| AAPL 10-K / AAPL 10-Q / FMCC section maps | — | byte-identical through the boundary fixes |
| Citigroup 10-K analyzer map | — | identical before/after |

- Section/TOC test battery (324 tests across the toc/section/corpus/guardrail files + #821/#836/gegs regressions): passes.
- New regression coverage: `tests/issues/regression/test_issue_904_toc_subset_overflow.py` (19 offline + 1 VCR e2e) and `test_issue_905_phantom_part_items.py` (7 offline + 1 VCR e2e), cassettes committed.
- Perf: the body scan only runs when the map is gappy; the successor scan's missing-number pre-gate means fully-anchored filings (AAPL) skip it entirely — no change to the #901 hot path.

## Commits

- `fix(sections): union-merge body-header items when the TOC anchors a subset (GH #904)`
- `fix(sections): skip same-anchor sections when resolving end boundaries (GH #904)`
- `feat(sections): flag schema-invalid part/item keys and embedded successor headers (GH #904, #905)`
- `test(sections): regression coverage for TOC-subset overflow and phantom 10-Q items (GH #904, #905)`

Changelog entries added under `[Unreleased]`.

## Post-review update

Both review findings are addressed; `tests/issues/regression/test_issue_891_split_node_item_metadata.py` (the NATH test that caught the blocker) is now part of the validation set and passes, and the section battery re-runs at 81 passed / 1 skipped.

- **Body-header maps collapsed onto stale anchors are rejected** (fewer than 80% distinct anchor ids). Filers with no per-item anchor divs (Nathan's Famous) stay on their pattern-detection path — fixes the issue-891 regression. The guard sits inside `_analyze_body_item_headers`, so it covers both consumers: the wholesale-replacement floor path and the union-merge. The union-merge additionally drops a recovered item whose anchor the TOC already claimed, and the logical-vs-positional boundary-ordering assumption is documented at the merge site.
- **The successor pre-gate treats optional Item 16 as no gap** (`FormSchema.optional_item_numbers`). Filings that omit it (JPM, XOM) are back to zero successor extractions; a map missing a mandatory trailing item (15) still opens the gate.

**Residuals on Coeur (surfaced, not fixed):** after the fix, two over-captures remain and are self-flagged by the successor guardrail at confidence 0.5 — `part_ii_item_8` embeds the Item 9 header and `part_iii_item_14` embeds the Item 16 header. This PR's claim for Coeur is "Item 7 and 1B repaired; remaining over-captures flagged instead of silent," not "fully fixed."

Post-review commits:

- `fix(sections): reject body-header maps collapsed onto stale anchors (GH #904, #891)`
- `fix(sections): treat optional Item 16 as no gap in successor pre-gate (GH #904)`
- `test(sections): regression coverage for PR #906 review findings (GH #904, #891)`


sf1tzp:
One design point worth explicit sign-off before this leaves draft: **the successor-header guardrail produces true-positive-but-noisy flags on stub absorption**.

On the Coeur filing itself, after the boundary repairs, two sections still get flagged and downgraded to 0.5:

- `part_ii_item_8` (133K) — its end anchor is Item 9A, so it absorbs the undetected **Item 9** stub (~100 chars of "None."). The flag is factually correct (the boundary does overshoot Item 9), but a 0.95→0.5 downgrade for absorbing a one-line stub is arguably disproportionate, and any filer whose TOC omits Item 9 while its body carries an `Item 9.` header would flag the same way.
- `part_iii_item_14` (16.6K) — absorbs Item 15/16 territory and the exhibit index; this one feels like a clearly-deserved flag.

Options I see:

1. **Keep as-is** — flag-and-return says wrong content should never be presented at full confidence, and callers who care can inspect `.warnings` to see the absorbed item is a stub. (This is what the PR currently does.)
2. **Warning without the confidence downgrade** when the absorbed successor is a "stub-class" item (9, 1B, 3, 4 — items that are routinely one-liners), reserving the 0.5 downgrade for absorbed substantial items.
3. **Raise `_SUCCESSOR_SCAN_MIN_CHARS`** or require the match to sit some distance before the section's end — cruder, but keeps the guardrail quiet on tail-stub absorption.

Happy to implement 2 or 3 if you'd rather keep 0.5 confidence reserved for content-level failures. Also flagging: the FMCC side of this PR is deliberately detection-only (per-part validity flags the phantom `part_i_item_5/6`); actually repairing the running-header mis-anchoring is left out of scope, so #905 shouldn't auto-close on merge.

sf1tzp:
Hey @dgunning! Long time user of edgartools here. I was reviewing some of the edgartools output in my project, and noticed a couple of parsing errors for certain companies. I created #904 and #905 from those that I saw.

Looking a bit closer, I identified a few more likely candidates to review. I uploaded this [list to a gist](https://gist.github.com/sf1tzp/5f458bf6359f2c4a736253a6329138ce) if you are interested in seeing the full list, but it may contain some false-positives as I'm still validating the list against v5.43.0's current behavior.

Before I start any automation or open any new Issues & PRs, I want to check with you and make sure that this PR, and the issue formats of #904 and #905  are acceptable, and that you would be open to receiving more like them.

VCR testing is very cool! I haven't seen it before. Please let me know if I'm adding cassettes the way you want them to be added. Should your CI generate these on a merge instead of trusting the PR author?

 Happy Friday, and thanks for edgartools

dgunning (changes requested):
Thanks for this — the diagnoses in #904 and #905 hold up line-for-line against the code, and the Coeur fix genuinely works: `Item 7` goes 257,440 → **91,682** chars, bounded at 7A with no embedded successor headers. The commit split is clean and the flag-and-return guardrails are well-argued.

One blocker, one perf issue, and a note on how the results read.

---

## 🔴 Blocker — the Nathan's Famous 10-K section map is destroyed

`tests/issues/regression/test_issue_891_split_node_item_metadata.py` passes on `main` (4 passed) and fails on this branch. Nathan's Famous FY2026 10-K (`0001437749-26-019923`):

|  | main | this PR |
|---|---|---|
| sections | 11 (`pattern`) | 23 (`toc`) |
| `mda` | 46,186 chars | **absent** |
| item spans | distinct, sensible | 1/1a/1b/1c all 149,318 · 2→9a all 57,291 · 11→16 all 132,255 |

### Mechanism

```
NATH: generic TOC canonical items = 0   (link-less TOC; floor = 8, so the gate is open on both branches)

main: body-scan items =  0  → count(body) > count(result) is False → no replacement
                            → pattern detection wins, produces the good 11-section map

PR:   body-scan items = 23  → 23 > 0 → wholesale replacement fires
      body map: 23 keys → only 4 DISTINCT anchors  (c47: 10 items, c61: 6, c31: 4, c50: 3)
```

The PR description says *"The wholesale-replacement floor path (GS/Citi class) is untouched."* That's the assumption that breaks — the floor path **consumes** `_analyze_body_item_headers`, and change 1 (bold-child recognition) changes what that returns. On NATH the looser bold check takes the body scan from 0 matches to 23, which trivially beats the empty TOC map, so the wholesale replacement now fires where it previously didn't.

The 23 matches are then bad: `_analyze_body_item_headers` resolves each header to `last_anchor_id`, and NATH doesn't emit an anchor div before each item header, so ten consecutive items inherit one stale id. Nothing checks anchor distinctness before the map is accepted, so every item in a run collapses onto the same span — and because a "complete" TOC map now exists, the correct pattern-detected `mda` is gone.

Reproduce:

```python
from edgar.documents.utils.toc_analyzer import TOCAnalyzer
a = TOCAnalyzer(form="10-K")
body = a._analyze_body_item_headers(html)          # NATH 0001437749-26-019923
print(len(body), len(set(body.values())))          # PR: 23 4   |   main: 0 0
```

### Suggested fix

Gate acceptance of a body map on anchor **distinctness**, not just key count — e.g. reject when `len(set(body.values())) < ~0.8 * len(body)`, or drop any item that reuses a preceding item's anchor. NATH then falls back to pattern detection exactly as on `main`, while Coeur (whose headers each have their own anchor div) is unaffected.

Please apply the same guard to the new union-merge. It has the identical latent exposure, and it now runs on a much wider population than the floor path did — `_CORE_TEN_K_ITEMS` includes `7A`, which smaller reporting companies may legitimately omit under Reg S-K 305(e), so any such filer triggers a body scan on every parse.

Separately, there's an unchecked assumption worth a comment at minimum: `result.update(recovered)` mixes TOC-link anchors with body-scan anchors, but `toc_section_extractor._analyze_sections` orders boundaries by *logical* item order (`_get_section_type_and_order`), not document position. A merged anchor that doesn't sit between its logical neighbours' positions can invert a boundary. The old wholesale-replace design avoided this structurally by never mixing sources.

---

## 🟡 Successor-guardrail pre-gate leaks whenever Item 16 is absent

The pre-gate requires every number in `own_num+1 .. max_item_number` (16) to be present, but Item 16 (Form 10-K Summary) is optional and commonly omitted:

| filing | Item 16 listed | successor `text()` extractions |
|---|---|---|
| AAPL / KO / CAT | yes | **0** — claim holds |
| JPM (items 1–15) | no | **5**, ~1.15M chars |
| XOM | no | 2 |

JPM section detection measured at **6.33s → 7.32s (+16%)**, best-of-3 with a warm HTML cache. So "fully-anchored filings pay zero extraction cost" holds only for filings that carry the optional item. Suggest capping `max_item_number` at the highest item number actually present in the map, or excluding optional items (16, 9B, 9C) from the required ladder.

---

## 🟢 Verified working

- Coeur `Item 7` bounded at 7A; `Item 1B` 317,604 → 4,474. Items 7A/8/9A/9B/9C/10/14 recovered.
- All 21 new regression tests pass; the offline/VCR split and `@pytest.mark.fast` usage match project conventions.
- Section battery (`test_10k_item_1b_1c`, `test_10q_section_detection`, `test_section_boundary_corpus`, `test_section_detection_edge_cases`, `test_section_detection_comparison`, `test_def14a_section_flip_gate`, `test_proxy_sections`): **81 passed, 1 skipped**.
- AAPL / KO / CAT: identical section maps, zero body scans, zero successor extractions, no false-positive warnings. The guardrails are not noisy.
- FMCC phantom flagging works as described, and scoping #905 as "Addresses" rather than "Fixes" is the right call — the mis-anchoring root cause stays open on our side.

## Note on the results table

After the fix, Coeur still shows two residual over-captures, correctly self-flagged by your new guardrail at confidence 0.5:

- `part_ii_item_8` embeds the Item 9 header
- `part_iii_item_14` embeds the Item 16 header

That's the guardrail doing its job and I'd rather have it than not — but the validation table currently reads as "Coeur fully fixed." Worth stating the residuals explicitly so nobody infers a stronger claim than the change makes.

Also FYI: the two size-band warnings on XOM are pre-existing on `main`, not from this PR.

---

Please add `tests/issues/regression/test_issue_891_split_node_item_metadata.py` to your validation set — it's the test that catches the blocker. Happy to re-review as soon as the distinctness guard is in.

sf1tzp:
Thanks for the thorough review — the NATH catch especially. All three points are
addressed; `test_issue_891_split_node_item_metadata.py` is now in the validation
set and passes 4/4.

## 🔴 Blocker — fixed via anchor-distinctness rejection

Your mechanism analysis was exactly right: the floor path consumes
`_analyze_body_item_headers`, and the looser bold check took NATH's body scan
from 0 to 23 matches — 23 keys collapsed onto 4 stale anchor ids.

I went with the whole-map rejection variant: `_analyze_body_item_headers` now
rejects its own result when fewer than 80% of items resolve to a distinct
anchor (`_MIN_DISTINCT_ANCHOR_SHARE = 0.8`). Placing the guard inside the
scanner covers **both** consumers — the wholesale-replacement floor path and
the union-merge — with one gate, including the wider population you flagged
(SRCs omitting Item 7A under Reg S-K 305(e) trigger a body scan on every parse;
a collapsed map can now never be accepted from either path).

I chose rejection over per-item drop deliberately: dropping reusers on NATH
would still leave a 4-item map that beats the empty TOC (`4 > 0`) and fires the
wholesale replacement — the pattern-detected map would still be lost. Rejection
restores `main` behavior exactly:

```
NATH: body scan → 23 items / 4 distinct anchors (0.17) → map rejected → {}
      → floor path no-op → pattern detection → 11 sections, mda = 46,186 chars
Coeur/GS/Citi: every header self-anchored (share 1.0) → unaffected
```

The threshold sits at 0.8 rather than ~1.0 because an *isolated* shared anchor
is legitimate — Coeur's Items 1B/1C share one anchor div (that's what the
same-anchor boundary fix handles); the guard targets collapse, not adjacency.
There's a boundary test pinning a 4-of-5-distinct map as accepted.

On the union-merge specifically, two additions:
- a recovered item whose anchor the TOC already claimed is dropped (two
  sections on one span otherwise);
- the logical-vs-positional ordering assumption you called out is now
  documented at the merge site. Agreed it's real — with per-item-distinct
  anchors now guaranteed upstream it's safe in practice, but if it ever bites
  I'd promote it to a positional sanity check rather than a comment.

## 🟡 Pre-gate leak — fixed via optional-item exclusion

Went with your second option (excluding optional items) rather than capping at
the highest item present: capping goes blind when *trailing mandatory* items
(13–15) are missing from the map, which is precisely an overflow signal.

`FormSchema` now declares `optional_item_numbers` (10-K: `(16,)`, per the
llmp.2 "schema as single home of form knowledge" convention) and the pre-gate
treats those numbers as no gap. 9B/9C need no entry — their *number* is covered
by the mandatory 9/9A ladder rung. JPM/XOM-class filings are back to zero
successor extractions; a map missing mandatory Item 15 still opens the gate
(both pinned by fast tests). Coeur's residual flags are unaffected — its map is
also missing mandatory numbers, so its sections still scan.

## Note on the results table — agreed

Updated the PR description to state the residuals explicitly: after the fix,
Coeur still has two self-flagged over-captures at confidence 0.5 —
`part_ii_item_8` embeds the Item 9 header and `part_iii_item_14` embeds the
Item 16 header. The claim is "Item 7/1B repaired + residual over-captures
surfaced instead of silent," not "Coeur fully fixed."

## Validation rerun (this branch, post-fix)

| suite | result |
|---|---|
| `test_issue_891_split_node_item_metadata.py` (NATH) | 4 passed |
| `test_issue_904` + `test_issue_905` (incl. 5 new review tests, Coeur/FMCC VCR) | 30 passed |
| `test_toc_agents.py` (GS/Citi body-header fixture class) | 44 passed |
| section battery (same seven files you ran) | 81 passed, 1 skipped |
| `test_section_size_guardrail.py` / form-schema suites | 9 / 32 passed |

dgunning (commented):
Re-reviewed at `5c19df89`. The blocker is genuinely fixed — I verified it against the filings rather than from the test results — and the anchor-distinctness approach is the right call. One residual on the perf item, and the cause is a mistake in my original review rather than in your fix.

---

## 🟢 Blocker — resolved, verified independently

Your repro snippet on this branch:

```
NATH body scan: 0 items, 0 distinct anchors     (was 23 / 4)
NATH sections: 11 | methods: {'pattern'} | mda: 46,186 chars
```

That is byte-identical to the `main` column of my earlier table. `test_issue_891_split_node_item_metadata.py`: 4 passed.

Putting the guard inside `_analyze_body_item_headers` rather than at the call sites was the better choice — one gate covers the floor path and the union-merge together, including the Reg S-K 305(e) population where a body scan fires on every parse. Your reasoning for whole-map rejection over per-item drop is right and worth keeping in the commit message: dropping reusers leaves NATH with a 4-item map that still beats the empty TOC and still fires the wholesale replacement, so the pattern-detected map would be lost either way.

The 0.8 threshold with the 1B/1C adjacency rationale is well-judged, and the boundary test pinning 4-of-5 as accepted is exactly the test I'd have asked for.

Coeur is unharmed: `part_ii_item_7` = 91,682 chars, `part_i_item_1b` = 4,474.

## 🟡 Perf — fixed for JPM, still leaking on XOM, and that one is on me

JPM is genuinely fixed:

| | successor `text()` extractions | detect time |
|---|---|---|
| JPM, `optional_item_numbers=(16,)` | **0** | 5.41s |
| JPM, field neutralised (pre-fix gate) | 5 | 6.20s |

XOM, however, still pays 2 — because **XOM lists Item 16 and is missing Item 6**:

```
XOM detected item numbers: [1,2,3,4,5,7,8,9,10,11,12,13,14,15,16]
missing mandatory: [6]
```

My review table asserted XOM had no Item 16. That was wrong; I read across from JPM without checking XOM's map separately. Item 6 has been "[Reserved]" since the 2021 amendments dropped Selected Financial Data, so filers omit it exactly as they omit 16 — it belongs in the same set. You implemented what I asked for correctly.

The correction is one token, and I verified it takes XOM to zero with nothing else changed:

```python
optional_item_numbers=(6, 16)
```

Scope is narrow — of {XOM, CVX, PG, WMT, KO}, only XOM is affected; the other four have complete ladders and were already at 0 before and after. Your `test_successor_pre_gate_zero_extractions_when_only_item_16_absent` is what pins the Item 16 case; a sibling asserting the same for a map missing only Item 6 would close it.

Measurement method, if you want to reproduce: fresh `HTMLParser(ParserConfig(form="10-K")).parse(html)` per run, counting `Section.text` calls across `HybridSectionDetector(...).detect_sections()`. Going through `filing.obj()` won't show it — the document is cached and section detection doesn't re-run.

## 🟢 Everything else

- Section battery (the same seven files): **81 passed, 1 skipped**.
- `test_issue_904` + `test_issue_905` + `test_toc_agents` + `test_section_size_guardrail`: 79 passed.
- The union-merge anchor-collision drop and the documented logical-vs-positional ordering assumption both land where I'd hoped. Agreed on promoting it to a positional sanity check only if it actually bites.
- Choosing optional-item exclusion over capping at the highest item present is the correct trade — capping does go blind precisely when trailing mandatory items are missing, which is the overflow signal.
- Residuals now stated explicitly in the description. That reads accurately.

---

Nothing else blocking from me. Happy to take this with the Item 6 line added, or to merge as-is and let it follow — your call which is tidier.

Separately, you asked about issue/PR format, the wider candidate list, and whether CI should generate cassettes rather than trusting PR authors. Those are @dgunning's calls and he'll come back to you on them directly — they deserve a proper answer rather than one buried in a code review.

dgunning:
Answering the three process questions from your earlier comment properly, since they deserve more than a line in a code review.

## Are #904/#905 and this PR the format you want? Yes.

Both issues had what makes a filing bug actionable: a specific accession number, the observed and expected character counts, and enough of a mechanism to check. I could reproduce each in a couple of minutes, and the diagnoses held line-for-line against the code — that is rare and it is most of the review cost gone. Keep writing them exactly like that.

The PR shape is right too: one commit per mechanism, validation table with before/after, regression coverage split into offline unit tests plus a VCR end-to-end. The thing I would keep doing above all is scoping the claim honestly — "Addresses #905" rather than "Fixes", and the residual over-captures stated rather than left for a reader to discover. That is what makes a table trustworthy.

## More of them? Yes.

I looked at your sweep gist. The pipeline is sound, and the instinct to do one representative filing per ticker first — because defects cluster by filer HTML template — is exactly right; it is the difference between fixing 65 bugs and fixing 6 with 65 confirmations.

File issues freely. Even the ones that turn out to be legacy damage in your store are useful signal about which templates are fragile.

For PRs, the one structural thing worth keeping is what you are already doing anyway: when several findings share a root cause, one PR fixing the cause with those filings as regression cases is much stronger than one PR each. It makes the fix's blast radius visible in a single diff, which is precisely what caught the Nathan's Famous regression here.

Your triage step already filters the "clean on live" cases before they reach me, which is the most valuable part of the whole pipeline. Please keep that gate in front of anything you send.

## VCR cassettes: your instinct was right, and no, CI should not generate them.

Recording in CI would put live SEC on the critical path of every merge — flaky and rate-limited — and, worse, it would *overwrite* rather than *check*. A cassette silently re-recorded on merge cannot fail, so the regressions cassettes exist to catch would be papered over at exactly the moment they mattered.

Where we are landing instead:

1. **Contributor-recorded cassettes stay the norm**, with the expectations written down in CONTRIBUTING (going in shortly): record from live SEC, never hand-edit a recorded response, record against `main` rather than against your fix, and keep them small by narrowing the test rather than trimming the file. Nothing in your cassettes here departs from that — I spot-checked Coeur `Item 7` against the live filing and got 91,682, matching your recording exactly.
2. **A repo check** — `hatch run check-cassettes`, plus a CI job ahead of the test jobs — verifying cassettes contain only plain recorded data. Written, landing on `main` shortly.
3. **Possibly later**, a periodic job that re-records a sample and diffs response bodies against the committed copies, reporting drift rather than overwriting. That catches staleness as filings and SEC responses change. Not built.

So: keep adding cassettes the way you did here. Mentioning in the PR description which filings you recorded and roughly when is helpful — it tells a reviewer what to spot-check.

Thanks for asking rather than assuming. That question was worth more than it looked.

dgunning (approved):
Thanks for your effort and the level of quality of this PR.

Approved

Issue #918, opened 2026-07-29
10-Q Part II Item 1 replaced by an MD&A cross-reference; numbered-index rows fabricate item anchors (ICE/CSCO/CTSH 10-Qs, FMCC 10-Ks)


sf1tzp:
## Issue Type
- [x] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.43.1
**Python Version:** 3.13.2
**Operating System:** Ubuntu (Linux 6.8.0)

## Bug Description
**What you expected to happen:**
`obj['PART II, Item 1']` on Intercontinental Exchange's Q3 2024 10-Q returns
the Legal Proceedings section — a legitimately short
incorporation-by-reference stub opening with its own heading:
`"ITEM 1. LEGAL PROCEEDINGS — See Note 13 to the consolidated financial
statements..."` (~144 chars).

**What actually happened:**
It returns **101,823 chars** opening *mid-sentence inside the MD&A's
cross-reference to the 10-K*: `'Item 1 "Business — Regulation" … included in
our 2023 Form 10-K …'` — mislabeled content from elsewhere in the filing, at
confidence 0.95 with no warnings in the section map, so there is no signal
to detect the defect without independently scanning the returned text.

Two distinct mechanisms produce this shape (both traced in 5.43.1):

1. **Cross-reference rescue** (ICE, CSCO, CTSH 10-Qs). The TOC anchor for
   Part II Item 1 is *correct*; the real stub is <200 chars, which trips the
   undersized-section rescue in `get_section_text`. The rescue's
   `_text_on_item_heading` check demands the item-number → title mapping of
   a 10-K (`_ITEM_TITLE_PATTERNS`: `1 → BUSINESS`) on a form where Item 1 is
   *Legal Proceedings*, so the correctly-anchored text is judged
   mis-anchored, and `_find_actual_item_content` then regex-hunts the raw
   HTML document-wide. The curly-quote entity `&#8220;` is composed entirely
   of separator-class characters, so the MD&A's `Item 1 "Business —
   Regulation"` cross-reference matches and silently replaces the section.
[...]

Issue #917, opened 2026-07-29
MCP server fails to start with mcp 2.0.0


xKDHx:
## Issue Type

- [ ] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [x] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.43.1
**Python Version:** 3.14.6
**Operating System:** macOS 26.5.2

## Bug Description
**What you expected to happen:**
`edgartools-mcp --test` should start successfully and report that the MCP server is ready to run.

**What actually happened:**
The MCP server fails to import with an AttributeError on startup. The edgartools[ai] extra declares mcp>=1.12.3, which allows uv/pip to resolve mcp 2.0.0. That release is a breaking change — it removed the decorator-based Server API (list_tools(), call_tool(), etc.) that edgar/ai/mcp/server.py relies on.

**Error message (if any):**
```
Traceback (most recent call last):
  File ".../bin/edgartools-mcp", line 6, in <module>
    from edgar.ai.mcp.server import main
  File ".../edgar/ai/mcp/__init__.py", line 21, in <module>
    from edgar.ai.mcp.server import main, test_server
  File ".../edgar/ai/mcp/server.py", line 140, in <module>
    @app.list_tools()
     ^^^^^^^^^^^^^^
AttributeError: 'Server' object has no attribute 'list_tools'
```

## Reproduction
**Company/Ticker:** N/A (MCP server startup issue)
**Time Period:** N/A
**Relevant Forms:** N/A

**Minimal code to reproduce:**
```bash
# Run via uvx — no pinning of mcp version:
uvx --from "edgartools[ai]" edgartools-mcp --test
```
Or equivalently:
```python
from edgar.ai.mcp.server import main  # raises AttributeError at import time
```

**Workaround**:
Pinning mcp below 2.0.0 resolves the issue:
```bash
uvx --from "edgartools[ai]" --with "mcp<2.0.0" edgartools-mcp --test
```

## Additional Context
[...]

PR #916, opened 2026-07-29
fix(sections): resolve Workiva TOC rows with split label/title hrefs (GH #915)

sf1tzp:
Fixes #915

## Problem

Tesla's FY2023 10-K (accession `0001628280-24-002390`) returned **145,561 chars** for `TenK['Item 3']` — the ~1.2K legal-proceedings body plus everything through Part II Item 5. Its Workiva TOC gives each row's "Item N." *label* link a different href than the *title* link, and the label hrefs are broken. `_analyze_workiva_toc` grouped a row's links by href and took the first group that parsed to an item — the label group:

- **Label target doesn't exist** → the row was dropped unless the title text happened to be in the keyword vocabulary. "Legal Proceedings" is, so Item 3 survived; "Unresolved Staff Comments", "Mine Safety Disclosures", and Item 5's long title are not, so Items 1B, 4, and Part II Item 5 silently vanished — and Item 3's end boundary overshot to the next detected anchor.
- **Label target exists but is wrong** → the item anchored there. Item 1's label points at the bare page-break div preceding Item 8's heading, so Items 1/1A landed in the wrong place and the knock-on boundary shift truncated Item 8 to 2.5K chars.

Same downstream shape as #904 (a missing successor anchor lets the previous section run long), but here the anchors go missing in the Workiva agent parser rather than the TOC-subset path.

## Fix

Each TOC row is now resolved as a whole instead of first-parsing-group-wins:

- Parse every href group first; when the row names a single item, choose its anchor across **all** of the row's href groups via a new `_choose_row_anchor` helper — preferring a target that exists and whose neighbourhood matches the item's own heading (reusing `_anchor_matches_heading`), falling back to row order.
- Rows where no group parses but the row text still reads "Item N. <title>" are recovered when exactly one group has a real target (extends the single-anchor row-text fallback to split title/page hrefs).
- Rows with multiple distinct items keep the historical per-group behaviour.

Healthy Workiva rows — label, title, and page number sharing one href — yield a single candidate and parse unchanged; the heading check only runs when a row has several distinct existing targets.

## Results on the reported filing

- The Workiva parse now returns all 24 sections with anchors **identical to the generic parser** (was 21, with Items 1/1A/6/7/7A/8 mis-anchored).
- `obj['Item 3']` → 1,165 chars, exactly the legal-proceedings body; every section at 0.95 confidence with no over-capture/truncation warnings.
- Item 1 grew from 10K → 45K chars, Item 8 from 2.5K → 161K chars.
[...]

PR #919, opened 2026-07-29
fix(sections): stop the short-section rescue rebuilding items from cross-references (GH #918)

sf1tzp:
First half of #918 — the issue stays open until the second half lands (draft PR #926).

## The defect — the short-section rescue rebuilds sections from cross-references (ICE/CSCO/CTSH)

For ICE's Q3 2024 10-Q (`0001571949-24-000017`), `obj['PART II, Item 1']` returned **101,823 chars opening mid-sentence** inside the MD&A's cross-reference to the 10-K ("Item 1 'Business — Regulation' … included in our 2023 Form 10-K"). The TOC anchor was **correct** — the failure is downstream in `SECSectionExtractor.get_section_text`:

1. The correctly-anchored Legal Proceedings stub is legitimately short ("See Note 13 …"), tripping the <200-char rescue.
2. `_text_on_item_heading` demanded the **10-K** title for the item number (`_ITEM_TITLE_PATTERNS: '1' → BUSINESS`) on a form where Item 1 is Legal Proceedings, so the correctly-anchored text was judged mis-anchored.
3. `_find_actual_item_content` then regex-hunted the **raw HTML document-wide** — and the curly-quote entity `&#8220;` is composed entirely of chars in the separator class `[\s&#;0-9xnbsp]`, so `Item 1 &#8220;Business…` in MD&A prose matched, and the section was rebuilt from another item's content.

## The fix

- A text that *opens* with the item's own "ITEM \<n\>" heading (tolerating ICE's "1(A)" paren style and a leading PART header line) counts as on-heading whatever the form — the title match stays as a fallback. Per review, a `(?<![0-9])` lookbehind keeps item 1 from matching the tail of "ITEM 11" (the separator class carries `0-9` for entity noise, so it would otherwise swallow the repeated leading digit) while `ITEM&#160;1.&#160;BUSINESS` still matches — both pinned by unit tests.
- The rescue's search is bounded to the window between the section's start and end anchors, since its premise (an anchor that landed on a PART header just before the body) means the real heading is never behind the anchor.

All three filers now return their real incorporation-by-reference stubs opening with "ITEM 1. LEGAL PROCEEDINGS" (the undersized-section guardrail still flags them, as it should).

## Verification

- `tests/issues/regression/test_issue_918_cross_reference_anchoring.py`: 10 unit tests on an ICE-shaped 10-Q (no network) covering the on-heading check (10-Q titles, paren style, PART prefix, no-title-entry items, wrong-number rejection, the Item 1 / ITEM 11 lookbehind) and the bounded rescue — plus a VCR-backed end-to-end test pinned to the ICE filing (144 chars opening on its own heading).
- **Section-map diff across the 55 committed 10-K/10-Q fixtures** (per review suggestion): **no section added or removed on any fixture**; ten short sections change content, all the same improvement class as ICE — Item 16 "None." stubs (ABBV, GOOGL, IBM, META, NVDA, PFE, PG, WMT, XOM) and AAPL's Item 11 stub now open on their own headings instead of rescue-fetched prose from other items (PG's Item 16 was 14K chars of Item 1 Business; WMT's was 25K chars, and its false size-guardrail downgrade clears to full confidence).
- TOC and section suites pass (245 tests), including all of `test_toc_analyzer_form_aware_bare_item.py`.

## Defect 2 (FMCC numbered-index rows) — split out per review

The bare-number corroboration approach is fully reverted (`toc_analyzer.py` and the #905 test/cassette now match `main`): as found in review, requiring positive corroboration from the link target silently dropped real items on filers whose anchor targets never sit near their headings (MS 10-K: 19 item sections → 6). The rework — recognising the numbered table/figure index by its header row instead of validating every bare-number row — is up as draft PR #926, which also fixes the standalone-replay gap in the #905 cassette.


dgunning:
Thanks for this — and before anything else: **the red `test-fast` was never yours.** `mcp` 2.0.0 shipped while your PRs were open and removed the decorator-based `Server` API, so `tests/test_mcp_server.py` failed at *collection*, turning `main` and every open PR red simultaneously. Fixed on `main` (162e5ffa, GH #917).

I've reviewed this rebased onto current `main`. **Defect 1 is good work and I want it. Defect 2 has a problem that I think blocks it in its current form.**

## Defect 1 (ICE/CSCO/CTSH cross-reference rescue) — sound

The diagnosis is exactly right, including the detail that makes it nasty: `&#8220;` being composed entirely of separator-class characters is why the MD&A cross-reference matched. Bounding the rescue to the window between start and end anchors is the correct fix, and the "opens with its own ITEM heading counts whatever the form" relaxation is right — `_ITEM_TITLE_PATTERNS` carrying only 10-K titles was always going to misjudge 10-Q items.

One small defect in that hunk:

```python
_text_on_item_heading("ITEM 11. EXECUTIVE COMPENSATION", item_num="1")  # -> True
```

The separator class `[\s &#;0-9xnbsp]*` includes `0-9` (needed for `&#160;` noise), so it swallows the first `1` and the literal matches the second. The new `(?![0-9A-Za-z])` guard blocks a *trailing* digit but not a leading one. Only the repeated-digit case leaks, so item 1 vs Item 11 is the only reachable collision in the 1–16 space, and it fails conservatively — the rescue is skipped rather than running the document-wide hunt. Verified fix is one token, a `(?<![0-9])` before the number, which clears it while keeping `ITEM&#160;1.&#160;BUSINESS` matching.

## Defect 2 (bare-number corroboration) — silently deletes real items

This is the blocker. I diffed the section map across all 54 committed 10-K/10-Q fixtures, before and after:

| Fixture | item sections before → after |
|---|---|
| `ms/10k/ms-10-k-2025-02-21.html` | **19 → 6** |
| `axp/10k/axp-10-k-2025-02-07.html` | 22 → 19 |

Morgan Stanley's 10-K loses `part_i_item_1c`, `part_i_item_3`, `part_i_item_4`, `part_ii_item_7a`, `part_ii_item_9/9a/9b/9c` and **all of Part III (10–14)**; the map falls back to generic name keys (`business`, `mda`, `risk_factors`…). So `Item 7A` (market risk), `Item 9A` (controls and procedures) and the entire executive-compensation block stop resolving on a major filer.

I isolated it to this hunk by monkeypatching `_bare_number_label_corroborated` to return `True` — the losses vanish completely, so it isn't interacting with anything else in the PR.

I then tried the two repairs I'd expect to work, and **neither does**:

1. **Treating a missing target as inconclusive** (reject only when the target exists and demonstrably isn't the heading) — no change at all. So this isn't the dangling-href case.
2. **Widening the `following::` lookahead** — I swept 3 / 5 / 8 / 12 / 20 elements. Identical results at every width.

The heading text just isn't reachable from the anchor target on these filers, so requiring positive corroboration rejects every bare-number row they have.

**Suggested direction:** narrow the guard by *context* rather than by anchor validation — detect that the row belongs to a numbered table/figure index (FMCC's "List of Tables" is a recognisable structure) and skip label extraction there, instead of demanding corroboration from every bare-number row. That targets the actual FMCC failure without making correctness conditional on a signal many filers don't emit.

This also breaks 4 pre-existing tests in `tests/test_toc_analyzer_form_aware_bare_item.py` (`TestBareItemFilteringIsFormAware`: `ten_q_accepts_valid_item_6`, `ten_k_accepts_item_8`, `no_form_preserves_default_cap_15`, `letter_suffix_preserved`). Worth noting *which* ones: every **accept** case fails, every **reject** case still passes. I nearly wrote those off as fixtures too thin to model a legitimately stricter rule — the fixture href is dangling — but the MS/AXP numbers say they were the canary.

## Minor

`test_fmcc_phantom_items_are_prevented` fails standalone: it requests `full-index/2026/QTR1/form.gz`, which isn't in its cassette. I confirmed this reproduces on your **unrebased** branch, so it's neither the rebase nor anything on `main` — it's the caveat you documented. It only passes when something earlier in the session warmed the quarter-index cache, so it'd be worth recording the request or stubbing the lookup.

## Suggested path

Split Defect 1 out and I'll merge it as-is (with the `(?<![0-9])` tweak). Defect 2 is worth getting right rather than fast — the FMCC bug is real and your diagnosis of it is correct; it's the enforcement mechanism that's too strict.

For context on why I ran the fixture diff at all: your audit is the reason this cluster is being worked, and a fix that silently drops 13 sections from Morgan Stanley would have landed in exactly the defect class you were reporting. The 484-test suite didn't catch it. I'd suggest that section-map diff as a standard check for this area — happy to share the script.

#916 is merged/merging separately; #921 came back clean (479 passed, 0 fixtures changed) and is queued behind it.

sf1tzp:
Done — this PR is now Defect 1 only, rebased onto main. The `(?<![0-9])` is applied with a unit test pinning both the ITEM 11 rejection and `ITEM&#160;1.&#160;BUSINESS` still matching. Defect 2 is fully reverted — your read matches mine: the corroboration signal doesn't exist on filers like MS, so any guard built on it rejects their whole bare-number population. I've reworked it as header-context detection (a bare number is only suppressed when its table's header names the numbering, e.g. Table | Description | Page, with no Item column) on a separate branch — fixture diff is clean, MS keeps all 19 — and re-recorded the #905 cassette cold so it replays standalone. Ran your suggested section-map diff across the 55 fixtures for this PR too: no sections added or removed; ten short stubs (mostly Item 16, plus AAPL's Item 11) now open on their own headings instead of rescue-fetched prose. Happy to take the script and wire it in as a standing check.

sf1tzp:
Sorry, we rebased and updated the code in the same push (so changes from intial PR to current are not diff-able). I'm reviewing now

sf1tzp:
Seems good to me, assuming no new regressions are found by the full test suite. I split the 2nd half of the original / issue into it's own branch, will rebase & activate that PR when this one is resolved. 

There are still several more cases to work through on my list. I intend to continue working through those at a metered pace, balancing my other WIP and not overloading review capacity. Cheers

Issue #915, opened 2026-07-29
10-K Item 3 overflows past its end boundary: Workiva TOC rows with split label/title hrefs mis-anchor or drop items (Tesla)


sf1tzp:
## Issue Type

- [x] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.43.1
**Python Version:** 3.13.2
**Operating System:** Ubuntu 24.04 LTS (kernel 6.8.0-136-generic)

## Bug Description

**What you expected to happen:**

`part_i_item_3`'s end boundary should stop at the next item's line-anchored header, which is present in the document body (`ITEM 4. MINE SAFETY DISCLOSURES`, and `ITEM 6. [RESERVED]` at offset 3,946 of the returned slice). When the TOC parse fails to anchor an item (here Part I Item 4), body-header recovery should bound the previous section instead of letting it run to the next detected anchor — and the item lookup should not return an over-captured slice with no signal.

**What actually happened:**

For Tesla, Inc.'s FY2023 10-K, `obj['Item 3']` returns **145,561 chars** — the actual legal-proceedings body (~1.2K chars of rendered text) plus everything through the end of Part II Item 5. The section map is missing `part_i_item_1b`, `part_i_item_4`, and `part_ii_item_5`, and anchors Items 1 and 1A at the wrong positions (with a knock-on shift that truncates Item 8 to 2.5K chars), so Item 3's end boundary overshoots to the next detected anchor.

Unlike older versions, the section map *does* flag this (`part_i_item_3` at confidence 0.5 with an explicit over-capture warning) — but the `obj[...]` item lookup surfaces none of that, so consumers using the item API get the over-captured slice with no signal on the returned value.

**Error message (if any):**
```
None — the item lookup is silent; the only signal is Section.warnings on the
section map (part_i_item_3 confidence 0.5, over-capture warning; part_ii_item_8
confidence 0.5, truncation warning).
```

## Reproduction
**Company/Ticker:** Tesla, Inc. (TSLA), CIK 0001318605
**Time Period:** FY2023 (period 2023-12-31, filed 2024-01-29)
[...]

PR #921, opened 2026-07-29
fix(sections): split colliding item anchors so Item 7A stops duplicating Item 7 (GH #920)

### LINKED ISSUE #920: 10-K Item 7A silently duplicates Item 7: page-number-only TOC collides two items on one anchor (Regions Financial)
## Issue Type

## Environment
**EdgarTools Version:** 5.43.1
**Python Version:** 3.13.2
**Operating System:** Ubuntu 24.04 LTS (kernel 6.8.0-136-generic)

## Bug Description

`obj['Item 7A']` should return Item 7A's own content — here a short incorporation-by-reference stub ("This information is set forth in the Risk Management section of Item 7 and is incorporated herein by reference.") — not Item 7's 194K-char MD&A body. Two distinct items should never resolve to the identical slice; if the anchors cannot be separated, returning a short stub, `None`, or a low-confidence warning would all be better than silently duplicating Item 7.

**What actually happened:**

For REGIONS FINANCIAL CORP's FY2021 10-K, `obj['Item 7A']` returns **194,033 chars** that open with `"Item 7. Management's Discussion and Analysis of Financial Condition and Results of Operations"` — the Item 7A lookup resolves to Item 7's content, and `obj['Item 7A'] == obj['Item 7']` is literally `True`.

The TOC pass assigns `part_ii_item_7` and `part_ii_item_7a` the **same anchor** (both start `..._49`, end `..._271`), so the two sections are the identical slice. Both appear in the section map at 194,033 chars, confidence 0.95, with no warning — so there is no signal to detect the defect without independently scanning the returned text.

**Error message (if any):**
```
None — the item lookup is silent. Both part_ii_item_7 and part_ii_item_7a
report 194,033 chars at confidence 0.95 with empty warnings.
```

## Reproduction
**Company/Ticker:** REGIONS FINANCIAL CORP (RF), CIK 0001281761
**Time Period:** FY2021 (period 2021-12-31, filed 2022-02-24)
**Relevant Forms:** 10-K, accession 0001281761-22-000016

```python
import os
import re

os.environ.setdefault("EDGAR_IDENTITY", "you your.email@example.com")
from edgar import find

f = find("0001281761-22-000016")
obj = f.obj()
print("full text:", len(f.text()))

content = obj['Item 7A'] or ""
print("Item 7A:", len(content), "chars")
print("head:", repr(content[:200]))
print("tail:", repr(content[-200:]))
print("same object as Item 7:", content == (obj['Item 7'] or ""))

# Line-anchored headers of items that should FOLLOW this section. Any hit
# means the extraction overran its end boundary. \s+ between "Item" and the
# number: filer templates often use a non-breaking space ("Item\xa07A.").
for item in ['Item 8', 'Item 9', 'Item 9A', 'Item 9B', 'Item 9C', 'Item 10']:
    pat = re.escape(item).replace("\\ ", "\\s+")
    m = re.search(rf"(?m)^\s*{pat}\s*[.:–—-]", content)
    print(item, "->", m.start() if m else None)

# Section map: name, chars, confidence, detection method, warnings
for name, s in obj.document.sections.items():
    txt = s.text() or ""
    print(name, len(txt), getattr(s, "confidence", None),
          getattr(s, "detection_method", None), getattr(s, "warnings", None))
```

Observed:

```
full text: 952,729

obj['Item 7A']: 194,033 chars
  head: 'Table of Contents \n\nItem\xa07.  Management’s Discussion and Analysis of
         Financial Condition and Results of Operations\n\nEXECUTIVE OVERVIEW\n\n...'
  tail: '...e and Qualitative Disclosures about Market Risk\n\nThis information is
         set forth in the Risk Management section of Item 7 and is incorporated
         herein by reference.'
  obj['Item 7A'] == obj['Item 7']  ->  True

section map (name: chars, confidence, method, warnings):
  part_ii_item_6:        19  0.95  toc  []
  part_ii_item_7:   194,033  0.95  toc  []
  part_ii_item_7a:  194,033  0.95  toc  []      <- identical slice to Item 7
  part_ii_item_8:   284,727  0.95  toc  []
  (all other items normal)
```

The real Item 7A is the two-line incorporation-by-reference stub visible at the tail of the returned slice.

### Root cause (generic TOC parser, page-number-only links)

This filing is **not** parsed by the Workiva agent path (`filing_agent` is `None`), so it goes through the *generic* TOC parser in `edgar/documents/utils/toc_analyzer.py`. Its linked TOC carries **only page numbers** — the "Item N." labels and titles are plain, non-link cells — so each item resolves via `_extract_preceding_item_label` to the anchor of the page it starts on. Item 7 (MD&A) and Item 7A both begin on **page 41**, and both page-number links point at that one page anchor (`..._49`):

```python
from edgar.documents.utils.toc_analyzer import TOCAnalyzer
a = TOCAnalyzer(form='10-K')
m = a.analyze_toc_structure(html)         # part_ii_item_7  -> ..._49
                                          # part_ii_item_7a -> ..._49  (collision)
```

Downstream (`toc_section_extractor._analyze_sections`) two distinct item keys on one anchor slice to the identical span. Note the shared-anchor handling added for #904 keeps both sections but gives them identical spans with no confidence penalty or warning.

The document body carries each item behind its own heading anchor — a body-header scan resolves `part_ii_item_7a` to a distinct anchor (`..._3742`) that sits correctly between Item 7's `..._49` and Item 8's `..._271`. That is enough to separate them.

### Proposed fix

A fork branch [`sf1tzp/edgartools@fix-regions-item-7a`](https://github.com/sf1tzp/edgartools/tree/fix-regions-item-7a) adds an anchor-collision repair pass at the end of `analyze_toc_structure`: when two distinct item keys resolve to a single anchor, re-resolve the displaced item from the body-header scan (each item's own heading behind its own anchor), guarded so the new anchor sits after the shared anchor and before the next item in document order (no boundary inversion). The item whose body anchor equals the shared anchor keeps it (MD&A owns the page it starts on); the other is re-pointed. When the body scan cannot separate them, the collision is logged rather than left silent.

Result for RF: Item 7A → its 181-char incorporation-by-reference stub, Item 7 → intact 193,850-char MD&A, no overrun into Item 8+. Verified offline against a committed RF fixture; 89 existing section/TOC tests pass with no regressions (including the #904 Coeur shared-anchor and #891 Nathan's Famous stale-anchor tests).

### Impact

Downstream consumers get mislabeled content from elsewhere in the filing, which silently duplicates content across sections and inflates token counts for NLP/RAG pipelines. Because both sections report at confidence 0.95 with no warnings, the defect is invisible without independently scanning the returned text.

A bulk scan of Fortune-500 filings (edgartools 5.43.1) surfaced the same "Item 7A returns the Item 7 MD&A body" shape on **SO** (10-K ×5) and **MCD** (10-K) as well. SO is the *same* root cause (generic-parser page-number anchor collision, Item 7 + 7A → one anchor), but its body item headings aren't matched by the current body-header scan (`_is_bold_header` / the body item-header pattern), so the proposed fix logs the collision but cannot yet recover it — broadening the body-header scan to those heading styles would extend coverage.

### Related prior art

- #904 (fixed in #906, 5.43.1) — the shared-anchor handling referenced above was added there for adjacent one-line stubs (Coeur Mining Item 1B/1C); it keeps both colliding sections but assigns them identical spans, which is what surfaces here as a silent duplicate on a large item.
- #915 / PR #916 — same *symptom* family (item boundaries wrong) but a different root cause: the Workiva-agent TOC parser's split label/title hrefs. RF goes through the generic parser instead, so #916 does not address it (verified: applying that branch leaves RF's `Item 7A == Item 7` unchanged).

---

sf1tzp:
Fixes #920.

## Problem

For REGIONS FINANCIAL CORP's FY2021 10-K (`0001281761-22-000016`), `obj['Item 7A']` returns Item 7's 194K-char MD&A body — `obj['Item 7A'] == obj['Item 7']` is literally `True`, both at confidence 0.95 with no warning.

## Root cause

This filing is **not** a Workiva-agent parse (`filing_agent` is `None`) — it goes through the *generic* TOC parser. Its linked TOC carries **only page numbers** (the "Item N." labels/titles are plain, non-link cells), so each item resolves to the anchor of the page it starts on. Item 7 (MD&A) and Item 7A both begin on **page 41** and collide on that one page anchor (`..._49`). Downstream, two distinct item keys on one anchor slice to the identical span. The shared-anchor handling added for #904 keeps both sections but assigns them identical spans with no confidence penalty or warning.

## Fix

`analyze_toc_structure` gains an anchor-collision repair pass: when two distinct item keys resolve to a single anchor, re-resolve the displaced item from the body-header scan (each item's own heading behind its own anchor), guarded so the new anchor sits after the shared anchor and before the next item in document order (no boundary inversion). The item whose body anchor equals the shared anchor keeps it (MD&A owns the page it starts on); the other is re-pointed. When the body scan can't separate them, the collision is **logged** rather than left silent.

For RF: Item 7A → its 181-char incorporation-by-reference stub, Item 7 → intact 193,850-char MD&A, no overrun into Item 8+.

## Verification

- New offline regression test (`tests/issues/regression/test_issue_920_rf_item7a_anchor_collision.py`) against a committed RF fixture — asserts 7A is its own stub, 7 stays intact, and 7A doesn't overrun.
- **89 existing section/TOC tests pass, no regressions** — including the #904 Coeur shared-anchor and #891 Nathan's Famous stale-anchor tests (the cases most at risk from touching anchor handling).

## Scope

This addresses the single-anchor collision class. A Fortune-500 scan surfaced the same "Item 7A returns the Item 7 MD&A body" symptom on **SO** (×5) and **MCD**. SO shares the *exact* root cause but its body item headings aren't matched by the current body-header scan (`_is_bold_header` / body item-header pattern), so this fix logs the collision but can't yet recover SO — broadening the body-header scan to those heading styles would extend coverage. Verified SO/MCD are byte-identical on this branch (no regressions). The other cluster shapes (CPNG Item 11, INTU/MCD Item 10, TSLA 10-Q Item 4) are different mechanisms, out of scope here.

## Relation to #915 / #916

Same symptom *family* (item boundaries wrong) but a different root cause — #916 fixes the Workiva-agent split label/title hrefs; RF goes through the generic parser, so #916 leaves `Item 7A == Item 7` unchanged (verified). This PR is independent and based on 5.43.1.


dgunning:
Reviewed and verified — merging. I've rebased this onto current `main` for you; the patch to `edgar/` is byte-identical to what you submitted (only the base blob hashes moved, since #916 landed in the same file). The CHANGELOG needed a manual merge where your two commits both touched the RF entry.

**The red `test-fast` was never yours** — `mcp` 2.0.0 shipped while your PRs were open and removed the decorator-based `Server` API, so `tests/test_mcp_server.py` failed at *collection*, turning `main` and every open PR red at once. Fixed on `main` (162e5ffa, GH #917).

Verification, re-run against `main` **after** #916 merged (since that PR changes anchor resolution in the same file):

- Offline TOC/section sweep: **484 passed, 2 skipped, 0 failures**
- Your regression tests and #916's, run together: **9 passed** — the two changes coexist
- Section-map diff across all 54 committed 10-K/10-Q fixtures: **0 changed, 0 lost, 0 gained** — the collision-repair path is inert on normal filings, exactly as intended

I checked the ordering assumption the fix rests on: `_get_section_type_and_order` gives `part_ii_item_7` → 207000 and `part_ii_item_7a` → 207001, so `keys_sorted[0]` is Item 7 and the fallback owner is the MD&A, as your docstring says.

One thing I raised against this PR and then withdrew, in case it's useful: it adds the file's first `logger.warning` (currently 13 `debug` / 2 `info` / 0 `warning`), plus a `logger.info` per successful repair. Given we've just come off a 7M-filing bulk-crawl bug report, per-filing log noise was a live concern. I measured it rather than guessing — **zero INFO/WARNING records emitted across 55 fixtures** — so it isn't a real problem, and the diagnosability you get from logging an unrecoverable collision is worth more than the theoretical noise. Good call on your part.

Your stated scope limit stands and I've recorded it: Southern Co and McDonald's share this root cause but their body item headings aren't matched by `_is_bold_header`, so the collision is logged rather than recovered. `edgartools-llmp.6.3` stays open for those.

This ships in 5.44.0. Thank you.

dgunning (approved):
Verified against main-with-#916: 484 passed / 0 failed on the offline TOC/section sweep, 9/9 with #916's regressions run together, and 0 of 54 fixtures changed. Ordering assumption (item_7 before item_7a) confirmed directly.

PR #982, opened 2026-08-06
perf: try fromisoformat before strptime in _parse_date (~15% of get_facts)

### LINKED ISSUE #981: [PERFORMANCE] _parse_date tries strptime before fromisoformat; SEC dates are ISO (~37x slower per call, ~15% of get_facts)
## Performance Issue Details

**Issue Type:**
- [x] Slow operation (longer than expected)
- [ ] Memory usage too high
- [ ] Memory leak (increasing memory over time)
- [x] CPU usage too high
- [ ] Hangs/freezes during operation

`EntityFactsParser._parse_date` tries `datetime.strptime` **before** `datetime.fromisoformat`. SEC facts dates are ISO-8601, so the first format in the loop always matches and `fromisoformat` — already present as the final fallback — is never reached. `strptime` goes through `_strptime` and takes a locale lock on every call; `fromisoformat` is a C fast path.

## Environment

**EdgarTools Version:** 5.45.1
**Python Version:** 3.13.2
**Available RAM:** 64GB
**CPU:** Intel i9

## Performance Metrics

**Expected performance:** date parsing should be a negligible fraction of `get_facts()`, well behind the network read.

**Actual performance:** date parsing is the **second largest cost** in `get_facts()`, at roughly **15% of wall-clock**, only behind the SSL read.

Per-call microbenchmark (200k iterations on `"2024-03-31"`, CPython 3.13.2):

```
strptime       3.60 us/call
fromisoformat  0.10 us/call
-> 37.7x faster
```

## Reproduction Details

**Company/Ticker:** WMT, HD, MCD, NKE, LOW, TGT (6 companies, 139,284 facts)
**Dataset Size:** full company facts per company
**Time Period:** all available

**Code to reproduce:**

```python
import cProfile, io, pstats, time

SYMBOLS = ["WMT", "HD", "MCD", "NKE", "LOW", "TGT"]

pr = cProfile.Profile()
pr.enable()
total = 0
for s in SYMBOLS:
    f = Company(s).get_facts()
    total += len(f.query().execute()) if f is not None else 0
pr.disable()
print(f"{len(SYMBOLS)} companies, {total} facts")

buf = io.StringIO()
pstats.Stats(pr, stream=buf).sort_stats("tottime").print_stats(10)
print(buf.getvalue())
```

Top of the profile (`tottime`):

```
ncalls    tottime  function
   172     2.288   {method 'read' of '_ssl._SSLSocket'}      <- network, for scale
364523     1.694   _strptime.py:413(_strptime)
139284     0.911   edgar/entity/parser.py:145(_parse_single_fact)
364523     0.252   {built-in method _locale.setlocale}
364523     0.250   {built-in method strptime}
```

**364,523 date parses for 6 companies**, with `_locale.setlocale` called once per parse.

Discounting cProfile's per-call inflation (measured unprofiled), those parses are about **1.3s of ~8.7s** for the 6 companies.

## System Resource Usage

- CPU usage: single core saturated during parsing (the work is CPU-bound, not I/O-bound, once bytes are in the HTTP cache)
- Memory usage: not an issue here
- Disk I/O: negligible
- Network usage: 2.29s of the 8.7s (the `_ssl.read` above)

## Comparison/Baseline

**Does this operation work normally with:**
- [x] Smaller datasets — cost scales linearly with fact count, so it is only visible at scale
- [ ] Different time periods
- [x] Different companies — worse for companies with more facts (JPM: 52,370 facts)
- [ ] Different forms

Not company-specific or form-specific: it is proportional to the number of facts parsed.

## Suggested fix

Try the ISO fast path first, keeping the existing formats as fallback:

```python
@staticmethod
def _parse_date(date_str: Optional[str]) -> Optional[date]:
    """Parse date string to date object"""
    if not date_str:
        return None

    # SEC facts dates are ISO-8601; fromisoformat is a C fast path (~37x faster
    # than strptime, which takes a locale lock per call).
    try:
        return datetime.fromisoformat(date_str).date()
    except ValueError:
        pass

    for fmt in ('%Y-%m-%d', '%Y%m%d', '%m/%d/%Y'):
        try:
            return datetime.strptime(date_str, fmt).date()
        except ValueError:
            continue
    return None
```

Behaviour notes (verified on CPython 3.13.2):

```
fromisoformat('2024-03-31')          -> 2024-03-31
fromisoformat('20240331')            -> 2024-03-31
fromisoformat('2024-03-31T00:00:00') -> 2024-03-31
fromisoformat('03/31/2024')          -> ValueError
```

So on CPython 3.11+ the fast path also covers `%Y%m%d`; only `%m/%d/%Y` still needs `strptime`. The `.date()` call normalises a `datetime`-shaped string exactly as the current last-resort branch already does.

## Additional Context

- **Is this a regression?** No — the ordering appears to be long-standing.
- **Patterns noticed:** the cost is entirely in the *first* `get_facts()` per ticker per process. `EntityFacts` is cached (I saw a second call for the same ticker return in ~0ms with 26 function calls), so this does not compound within a run — but it does dominate any workload that touches many distinct tickers once each.
- **Workarounds found:** none that avoid the parser; enabling local storage does not help, since the cost is parsing rather than downloading.
- Found while profiling a nightly pipeline that calls `get_facts()` across several hundred symbols.

Happy to send a PR with the change plus a small benchmark if that is useful — just say the word.

joseturegano:
Closes #981.

## What

`EntityFactsParser._parse_date` tried `datetime.strptime` before `datetime.fromisoformat`. SEC facts dates are ISO-8601, so the first format in the loop always matched and the `fromisoformat` call already present as the final fallback was never reached.

`strptime` goes through `_strptime` and takes a locale lock on every call; `fromisoformat` is a C fast path.

## Measurement

CPython 3.13.2, edgartools 5.45.1, Windows 11.

```
strptime       3.60 us/call
fromisoformat  0.10 us/call        -> 37.7x
```

`cProfile` over `Company(sym).get_facts()` for WMT, HD, MCD, NKE, LOW, TGT (139,284 facts):

```
ncalls    tottime  function
   172     2.288   {method 'read' of '_ssl._SSLSocket'}     <- network, for scale
364523     1.694   _strptime.py:413(_strptime)
139284     0.911   edgar/entity/parser.py:145(_parse_single_fact)
364523     0.252   {built-in method _locale.setlocale}
364523     0.250   {built-in method strptime}
```

364,523 date parses for 6 companies, with `_locale.setlocale` called once per parse. Discounting cProfile's per-call inflation (re-measured unprofiled), that is roughly **1.3s of ~8.7s — about 15%**, second only to the SSL read.

## Correctness

The change is a reordering, so the test pins **equivalence**, not speed: 14 input shapes checked against the pre-change implementation used as the oracle.

Two cases are worth calling out because they are the ones a reordering could plausibly break:

- **`'03/31/2024'`** is not ISO, so it is only reachable through the `strptime` loop. The test asserts it still parses — if someone later removes that loop as dead code, this fails.
- **Non-string input** (`12345`): the old `except Exception` swallowed the `TypeError` that `fromisoformat` raises. The new code catches `(ValueError, TypeError)` explicitly, so it still returns `None` rather than propagating.

Also covered: `'  2024-03-31  '` (padded — rejected before, still rejected), ISO datetimes with and without offset, `'2024-13-45'`, empty string and `None`.

**0 differences across all 14 cases.** 17 tests pass.

## Notes

- On CPython 3.11+ `fromisoformat` also accepts the bare `%Y%m%d` form, so only `%m/%d/%Y` still needs `strptime` — the loop is kept for exactly that.
- `EntityFacts` is cached per process, so the cost is entirely in the first `get_facts()` per ticker. This does not compound within a run, but it dominates any workload touching many distinct tickers once each — which is how I ran into it, profiling a nightly pipeline over several hundred symbols.
- Changelog entry added under `## [Unreleased]` / `### Performance`.


dgunning (commented):
Thanks for this — the diagnosis is right, and the way you evidenced it is better than most changes that land here.

I verified the change independently rather than taking the PR at its word, so here is what I checked and what I'd like changed before merge.

## Verified

**Correctness — no divergence.** I probed 15 input shapes beyond your 14, picked specifically to break a `fromisoformat`-first reordering: `Z` suffix, ISO week dates (`2024-W13-7`), ordinal dates (`2024-092`), compact datetimes (`20240331T000000`), space separator, microseconds, `0001-01-01` / `9999-12-31`, unpadded `2024-2-3`, 2-digit years, `bytes`, `list`, and `nan`. Zero divergences against the pre-change implementation.

The reordering is safe precisely *because* you kept the strptime loop, which is also what preserves behaviour on Python 3.10 where `fromisoformat` is much stricter than on 3.11+. You reasoned about that explicitly in the PR body and got it right — worth noting that our CI only runs 3.13 while the package claims `>=3.10`, so nothing here would have caught it if you hadn't.

**Performance — claim holds.** 22.7x on CPython 3.11.6 against your 37.7x on 3.13. Same order, same direction.

**Tests pass.** Your 17, plus 402 entity/facts/parser tests on the branch. The new file is correctly picked up by `test-fast`, so it won't fall through the marker classification hole we closed last week.

## Please change

**1. Drop the reformatting.** The substantive change is about 15 lines in `_parse_date`. The remainder of the +193/−102 is a single→double quote conversion, added trailing commas, and reflowed signatures across the whole of `edgar/entity/parser.py`.

Two reasons to split it out: `edgar/entity/` is predominantly *single*-quoted (3,589 literals against 1,366) and ruff enforces no quote rule here, so this moves against local convention rather than toward it; and it buries a 15-line perf fix inside a 295-line diff, which costs us `git blame` on those lines and makes the perf change impossible to revert on its own if it ever needs to be.

A `_parse_date`-only diff would be about 15 lines plus the test and changelog, and I'd merge it as-is.

**2. Give the exception narrowing its own changelog sentence.** Going from `except Exception` to `except (ValueError, TypeError)` is a real behaviour change and the body only frames it as "TypeError still returns None." The broader effect is that any *other* exception now propagates and aborts the whole company parse rather than yielding one `None` among 364,523 calls.

To be clear, I think that is the right direction — it is exactly what our unified-error-policy work wants, and I'd rather know than silently get a `None`. It just shouldn't be invisible inside a performance entry.

**3. Rename the test to `test_issue_981.py`.** Convention in `tests/issues/regression/` is `test_issue_NNN.py` — 235 of the 248 files there follow it, and it's in CONTRIBUTING.

**4. The changelog entry is 82 words against our 80-word cap.** Trivial, but the standard is new and I'd like to hold it from the start.

## Worth saying

The equivalence testing is the part I want to call out. Pinning behaviour rather than speed, keeping the pre-change implementation in the test file as an explicit oracle, and adding `test_non_iso_format_still_reaches_the_fallback` as a guard against a future maintainer deleting the strptime loop as "dead code" — that last test anticipates a failure mode that hasn't happened yet, which is not a common instinct. The measurement is properly attributed too: cProfile output with the SSL read left in for scale, and the per-call inflation discounted rather than quoted raw.

I've approved the workflow run, so CI will now execute — it hadn't run at all, since first-time fork contributions need a maintainer to release them.

dgunning:
Correction to my review above: I said our CI only runs 3.13. That's wrong — the matrix is `["3.10", "3.13"]`, narrowed to `["3.13"]` only for pull requests (`python-hatch-workflow.yml:78`). So 3.10 is tested, just not on your PR; a 3.10 break would be caught after merge rather than before.

Doesn't change anything about your change — the strptime loop you kept is what makes it safe on 3.10 either way — but I don't want to leave a wrong claim about the project's coverage standing in a review.

joseturegano:
Thanks — the review was more thorough than the change deserved, and one of the four points turned out to matter more than the other three. All are addressed in `50769732`.

**1. Reformatting dropped.** `parser.py` is back to `main` with only `_parse_date` touched: +15 −10, and zero quote changes. You were right that it moved against local convention — I hadn't noticed my editor had requoted the file, which is exactly the kind of thing that is invisible to the person who did it.

**2. The exception narrowing has its own entry**, under `### Changed` rather than a sentence inside the perf one. It reads better separated, since the behaviour change is not a performance claim.

**3. Renamed** to `test_issue_981.py`, and the module docstring now carries the issue URL that `tests/issues/regression/README.md` asks for.

**4. Both entries end with `(GH #981)`** and are within the cap — 70 and 50 words.

---

**The part worth flagging: the test file you praised did not test the bug.**

I restored the pre-#981 `_parse_date` and ran the file against it. All 17 tests passed. The equivalence oracle is genuine, but the pre-change implementation satisfies equivalence *by construction* — so the file guarded the reordering against changing behaviour and could not fail if someone reintroduced strptime-first. Your README asks for "the specific bug that was reported", and the reported bug is that the ISO fast path is never reached. That was the one thing the file did not assert.

`test_iso_input_never_reaches_strptime` now counts strptime calls through a `datetime` subclass. Verified both ways: it fails against `main`'s implementation (`assert 1 == 0`) and passes here. `test_non_iso_input_does_reach_strptime` is its positive half, so that deleting the fallback loop would not satisfy it either.

I mention it because you singled out the test design as the good part, and it had a hole in the middle. Better that it comes from me than from whoever "simplifies" `_parse_date` in two years and sees a green suite.

---

Noted on the CI matrix — thanks for correcting it rather than letting it stand. It does change one thing for me: since 3.10 is tested after merge and not on the PR, I was more careful about the strptime loop than I would otherwise have been, and I'd rather that reasoning be in the commit message than in my head.

One local note, take it or leave it: I couldn't run the full `tests/issues/regression/` suite on Windows. Installing per the pin (`vcrpy<8.2` → 8.1.1) gives `AttributeError: module 'aiohttp.streams' has no attribute 'AsyncStreamReaderMixin'` from `vcr/stubs/aiohttp_stubs.py:21` against the aiohttp that resolves today. I verified it is unrelated to this change — the failure set is byte-identical with the patch and with `parser.py` reverted to `main` (63 either way, zero differences) — so it is an environment issue on my side, not a regression. Flagging it only in case other contributors hit the same wall.

Issue #332, opened 2025-06-24
6-K Filings - Financials


firmai:
AttributeError: 'CurrentReport' object has no attribute 'financials'

Thanks,
Derek

dgunning:
## Issue Resolved ✅

### Problem
Users experienced  when trying to access financial data from 6-K filings.

### Root Cause  
The  class (used for 6-K and 8-K filings) was not inheriting from , which provides financial functionality.

### Solution Implemented
✅ **Made CurrentReport inherit from CompanyReport** in edgar/company_reports.py
✅ **Verified 6-K forms contain financial data** (confirmed with Novartis example)  
✅ **Confirmed Financials.extract() works on current reports**
✅ **Added comprehensive test coverage** to prevent regression

### What's Fixed
- ✅  property now accessible on 6-K filings
- ✅ , ,  properties available
- ✅ Maintains full compatibility with existing 8-K functionality  
- ✅ No breaking changes to existing code

### Code Changes
```python
# Before (broken)
class CurrentReport():
    
# After (fixed) 
class CurrentReport(CompanyReport):
```

### Testing
Added regression tests to prevent this issue from reoccurring:
- 
- 

### Commit
Fixed in commit: 6008c8f

The fix is now ready and 6-K filings will have full financial functionality! 🎉

Thanks for reporting this @firmai!

PR #1013, opened 2026-08-10
test: the reproduction scripts that swallowed their own failures (07lk.24 finding 3)

dgunning:
Closes the exception-swallower block of `edgartools-07lk.24` finding 3, for `tests/issues/reproductions/`.

## What was there

Twelve files carry `@pytest.mark.regression` and an `except` handler that discards the failure.

| | files | |
|---|---|---|
| Zero assertions — cannot fail under any input | **8** | deleted |
| Real assertions, localized swallowing | **4** | converted |

Two of the eight (`test_currency_fix.py`, `test_issue_411_solution.py`, 5 tests) were auto-marked `fast` by filename, so they ran inside the **required `test-fast (3.13)` gate** while calling the live SEC — and passed only because every failure became a print.

## Deleted (8)

Each is an exploratory debug script that grew a marker: `print`, `dir()`, `hasattr` probing, and in one case a hardcoded `/Users/dwight/PycharmProjects/edgartools` on `sys.path`. `test_6k_with_financials.py`'s only collected test was a run of `print()` calls describing a proposed design change. `test_issue_411_solution.py` finished by printing `✓ USER'S DESIRED WORKFLOW NOW WORKS!` unconditionally. Same species as `test_multiple_companies_429.py`, deleted in #1011.

**Checked before deleting** — every bug they touch is covered elsewhere by tests that assert:

| issue | covered by |
|---|---|
| 412 | `test_issue_412_historical_periods.py`, `test_issue_412_regression.py` |
| 332 | `test_issue_332_regression.py` |
| 429 | `test_issue_429_period_selection.py`, `test_issue_429_statement_period_regression.py` |
| 411 accessors | `get_revenue` 8 files, `get_net_income` 6, `get_total_assets` 3, `get_financial_metrics` 3, `get_operating_cash_flow` 1 |

## Converted (4)

**`test_integration_438_fix.py`** — rewritten. All three tests passed unconditionally: one returned early on missing filings, missing financials and a missing income statement; one wrapped its loop body in `except Exception: continue`, and since `AssertionError` is an `Exception` both of its asserts were dead; the third had no assertions at all. Now asserts invariants — deduplication never adds rows, never grows the revenue count, and is idempotent — rather than row counts, which move every time these companies file. Per-company failures are aggregated and reported together, so one bad filing does not hide the other two.

**`test_issue_332_6k_financials.py`** — the integration test is now explicitly `@pytest.mark.network` and asserts. It had three failure modes and all three were skips, including the `AttributeError` on `.financials` that issue #332 is entirely about.

**`test_412_regression.py`, `test_fix_438.py`** — dead `__main__` runners removed. Both caught `Exception` around every test, printed a `✗`, and exited 0, so running the file directly to check a fix reported success either way.

## Verification

Not trusted on a green run — each guard was checked against a planted offender:

| offender | result |
|---|---|
| deduplication adds a row | all 3 tests in `test_integration_438_fix.py` fail |
| `_is_revenue_concept` always False | NVDA test fails |
| 6-K lookup returns empty | 332 integration test fails |

Full fast suite: **4,579 passed**, 4 skipped, 1 xfailed. Collection guard in `tests/conftest.py` still passes (7,652 collected), so nothing was left unclassified.

## Also

Six now-dangling names dropped from `NETWORK_PATTERNS` in `tests/conftest.py`, including `test_multiple_companies_429` which had outlived its file since #1011. A pattern for a deleted file matches nothing, so it is not a bug — but it reads as coverage the list does not have.

`REPRODUCTION_FILES_BUG_ANALYSIS.md` is annotated rather than edited: it is a dated report, and its central finding (these files turn failures into silent passes) was right — the fix just turned out to be deletion rather than more error handling.


dgunning:
Pushed a ninth deletion, and it changes the conclusion of this PR.

After opening this, I audited the **whole** reproductions tree for zero-assertion files rather than only for swallowing `except` handlers. That surfaced `data-quality/test_currency_usd.py` — the twin of `test_currency_fix.py` already deleted here, AAPL/USD instead of Deutsche Bank/EUR, prints only, live SEC call.

**The scan I ran first could not have found it.** It has no `except` handler at all: it simply never asserts, so there is nothing to swallow. A scan for "tests that discard their failures" only finds tests that *have* a failure to discard. Scanning for "zero assertions" is the stronger query.

Superseded by `tests/issues/regression/test_issue_440_currency_display.py`, which asserts the behaviour properly — `test_us_company_usd_currency_display` pins `iso4217:USD` and requires `$` in the render, and `test_deutsche_bank_eur_currency_display` covers the EUR side.

### The tree is now finished, not merely improved

I had been treating this as a slice of a large backlog, on the strength of the bead's header figure ("136 .py files, only 54 collected"). That figure is from 2026-07-30 and the tree has been pruned hard since.

Measured directly on each tree — `git ls-tree` for the file counts, `pytest --collect-only` for the tests:

| | .py files | collected tests | files with zero assertions |
|---|---|---|---|
| `main` @ 67451372 | 17 | 51 | 9 |
| this branch @ 7c235a1a | 8 | 36 | **0** |

(Both file counts include `conftest.py`, so the test-file counts are 16 and 7.)

So this closes finding 3 for `tests/issues/reproductions/` outright, rather than leaving a remainder. The bead has been corrected.

*Edited: an earlier version of this comment said 25 files before / 8 after. The 25 was arithmetic on the bead's stale header figure rather than a measurement — the same mistake the comment is about. The counts above are measured.*

PR #1017, opened 2026-08-10
ci: report a red main on the lane that owns the coverage floor

dgunning:
#999 gave `regression-tests.yml` a reporter that opens and closes one labelled tracking issue when the post-merge lane goes red. It was written after that lane went red on 2026-08-08 and **three more pull requests merged through it before anyone noticed** (`edgartools-hwdp`).

`python-hatch-workflow.yml` never got one — and it carries more:

| job | what it owns | required? | watched? |
|---|---|---|---|
| Combine Coverage | **the 65% floor** (`coverage report --fail-under=65`) | no | **no** |
| test-network | the SEC integration suite | no | **no** |
| test-slow | the slow suite | no | **no** |

All three are post-merge only. Until now, a coverage regression below 65 landed in the Actions tab and nowhere else.

For reference, `main` is currently at **72%** (`TOTAL 68788 16609 28700 4301 72%` on `aa7ea7ec`), so there's ~7 points of headroom — this is about noticing when that's spent, not about being near the edge today.

## Why report rather than require

Making Combine Coverage a required check is not an option. It `needs` `test-network` and `test-slow`, which are **skipped** on pull requests, so a required Combine Coverage would never report and would block every PR forever — the deadlock documented at the top of the file. Reporting is the right mechanism for a check that cannot gate.

## Two details that matter

**`needs` names every post-merge job, not just coverage.** A `test-network` failure *skips* Combine Coverage rather than failing it, so keying on the coverage result alone would stay silent for the one failure that stops the threshold being evaluated at all.

**Its own label (`red-main-build`) and title**, separate from the regression workflow's `red-main`. Sharing one issue would have the two lanes fight over it — whichever went green last would close an issue the other still needed open.

The body also names *which* job failed. This lane has five, and "the build is red" without saying which one sends the reader to the Actions tab to find out, which is the step nobody takes.

## Verification

The red/green decision was checked against all nine result combinations that can actually occur:

| must report GREEN | must report RED |
|---|---|
| all success | **Combine Coverage failure** ← the case this exists for |
| coverage skipped | test-fast failure |
| all cancelled (superseded push) | test-network failure → coverage *skipped*, still red |
| PR-shaped skips | cassette gate failure → everything else skipped |
| | test-slow failure |

All nine behave correctly. The issue body was also rendered to confirm the literal `65%` survives `printf` — `65%%` in the format string, which is exactly the kind of thing that silently mangles a message.

Note this PR touches no `edgar/**` path, so the advisory regression lane correctly does not run on it.

### DIFF
--- a/.github/workflows/python-hatch-workflow.yml
+++ b/.github/workflows/python-hatch-workflow.yml
@@ -31,6 +31,14 @@
 # test-network, test-slow and Combine Coverage are deliberately NOT required:
 # they carry `if: github.event_name != 'pull_request'` and so report as skipped
 # on every pull request. They run on push to main, which is the integration gate.
+#
+# Because they cannot gate, they are REPORTED instead — see the `report` job at
+# the end of this file. Combine Coverage is where the 65% floor is enforced
+# (`coverage report --fail-under=65`), and until 2026-08-10 a failure there, or
+# in test-network or test-slow, appeared in the Actions tab and nowhere else.
+# That is the same hole #999 closed for the regression workflow after a red main
+# survived three merges unnoticed (bead edgartools-hwdp); this workflow was left
+# out of that fix even though it carries more.
 
 name: Build and Test Edgartools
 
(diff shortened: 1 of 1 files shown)


dgunning:
Force-pushed a rebase. The first push of this branch was wrong.

I branched it off `test/regression-provenance` (#1016) instead of `main`, so it opened carrying **120 files** — #1016's 119 provenance changes plus this one. The description said it touched no `edgar/**` path and that the advisory regression lane correctly would not run on it; both were false, and the regression lane duly triggered, which is what exposed the mistake.

Rebased onto `main` with `--onto`, so it is now **1 commit, 1 file** (`.github/workflows/python-hatch-workflow.yml`), and the regression lane no longer triggers — which is what the description claims and now what happens.

No content changed, only the base. #1016 is unaffected and still holds the provenance work.

PR #1021, opened 2026-08-10
fix(entity): require word boundaries for spaced LP suffix

### LINKED ISSUE #1019: `Entity` incorrectly classifies individual names containing "l p" as companies
## Issue Type

## Summary

`Entity` incorrectly classifies some individuals as companies. The issue is visible through edgartools' public `Entity` API: personal names are returned with `is_individual == False` and `is_company == True`.

The root cause appears to be the internal `_name_suggests_company()` heuristic in [`edgar/entity/constants.py`](https://github.com/dgunning/edgartools/blob/82180dc5327caf29c1b2951dc6e771165085f7da/edgar/entity/constants.py#L196-L226). It uses substring matching for the `"L P"` company-name keyword. This creates false positives when `"L P"` occurs naturally across word boundaries in a person's name. For example:

```python
_name_suggests_company("Michael P. Wyatt")
```

returns `True` because the uppercased name contains `"L P"` across the end of `"Michael"` and the `"P"` middle initial:

```text
MICHAEL P. WYATT
      ^^^
```

The loose substring check against the legal entity suffix unfortunately matches also ordinary personal names.

## Environment

- **edgartools version:** 5.43.0
- **Python:** 3.10.12
- **OS:** Ubuntu 22.04

## Reproduction

```python
import edgar
edgar.set_identity("you@example.com")
for cik in [1306539, 1564926, 1571320, 1785884, 2037927, 2097014]:
  e = edgar.Entity(cik)
  print(e.name, e.is_individual, e.is_company)
```

## Suggested fix

Change `"L P"` from a loose substring keyword to a strict/word-boundary match

nightcityblade:
## Summary

- treat spaced `L P` as a whole-word legal suffix instead of a loose substring
- preserve detection for names such as `ACME L P`
- add regression coverage for SEC entity 2037927 (`WYATT MICHAEL P.`)

Fixes #1019

## Testing

- `.venv/bin/python -m pytest -q tests/test_entity_core.py` — 52 passed
- focused Ruff checks on the changed source and test files — passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`Entity(2037927)` classified Wyatt Michael P. as a company.** The name heuristic treated spaced `L P` as a loose substring, matching across the end of "Michael" and the middle initial. Spaced `L P` now requires whole-word boundaries while remaining a recognized legal suffix. (GH #1019)
+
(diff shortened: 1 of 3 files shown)


dgunning:
Thanks for the patch. I've fixed this in #1027 — the same substring defect also affected `L L C` (`MICHAE|L L C|OOPER`), and dropping `L P` costs three `S.A.` companies their only name signal, so I took the fix wider than this diff. Closing in favour of that. Credit for the diagnosis goes to @mwtarnowski in #1019.

Issue #749, opened 2026-04-01
13F-HR: Value column units inconsistent — thousands (pre-Q4 2022) vs dollars (post-Q4 2022)


rustyconover:
## Bug

The `infotable` DataFrame's `Value` column has inconsistent units across filing periods:

- **Pre-Q4 2022**: Value is in **thousands** (as the SEC 13F XML schema specifies)
- **Post-Q4 2022**: Value is in **dollars** (the SEC changed the XML schema)

This means summing Value across time periods gives nonsensical results.

## Reproduction

```python
from edgar import Company, set_identity
set_identity("test test@test.com")
c = Company(1665241)  # Schonfeld Strategic Advisors

# Q3 2022: NVIDIA — value in thousands
f1 = c.get_filings(form="13F-HR", filing_date="2022-11-01:2022-11-30")[0]
info1 = f1.obj().infotable
nvda1 = info1[info1["Issuer"].str.contains("NVIDIA")].iloc[0]
print(f"Q3 2022: Value={nvda1['Value']:,}, Shares={nvda1['SharesPrnAmount']:,}")
# Value=9,204 Shares=75,820 → implied $0.12/share (value is in thousands: $9.2M)

# Q4 2022: NVIDIA — value in dollars
f2 = c.get_filings(form="13F-HR", filing_date="2023-02-01:2023-02-28")[0]
info2 = f2.obj().infotable
nvda2 = info2[info2["Issuer"].str.contains("NVIDIA")].iloc[0]
print(f"Q4 2022: Value={nvda2['Value']:,}, Shares={nvda2['SharesPrnAmount']:,}")
# Value=58,993,357 Shares=403,677 → implied $146.14/share (value is in dollars)
```

## Root cause

The XML parser in `edgar/thirteenf/parsers/infotable_xml.py` line 90 reads the raw `<value>` element as-is:

```python
info_table['Value'] = get_int("value")
```

The SEC changed the 13F XML namespace around Q4 2022. Older filings use a schema where `<value>` is in thousands; newer filings use a schema where `<value>` is in dollars.

## Suggested fix

Normalize all values to a consistent unit (dollars). Detect the schema version from the XML namespace and multiply by 1000 for older filings, or always report in thousands with clear documentation.

The transition point is between the 2022-09-30 and 2022-12-31 report periods.

dgunning:
## Triage

**Classification**: Bug — Data Accuracy  
**Severity**: High — silent 1000x value error for cross-period 13F analysis  
[...]

PR #1071, opened 2026-08-21
docs: correct stale $1000s guidance in holdings AI skill

hirekarl:
## Summary

`edgar/ai/skills/holdings/sharp-edges.yaml` and `skill.yaml` both still tell agents to manually multiply `Value` by 1000 to convert to dollars. That's been wrong since 5.28.2 (#749) normalized 13F values to whole dollars at parse time — following the current example code now overstates every value 1000x.

- **`value-in-thousands`** rewritten to reflect the 5.28.2 fix; `detection_pattern` now flags the (now-wrong) `* 1000` pattern instead of its absence
- **`skill.yaml`**'s two matching spots (the `Value ($1000s)` column comment and the `holdings['Value'].sum() * 1000` example) updated to match
- **`holdings-vs-infotable`** extended to cover a related gotcha found while digging into this: for multi-manager joint filings (e.g. Berkshire Hathaway's ~20 subsidiary managers), a filing's cover-page Entry Total counts disaggregated `infotable` rows, not aggregated `holdings` rows — comparing `holdings`' row count against Entry Total will always look short even when parsing is perfect, since `holdings` intentionally collapses manager combinations into one row per security
- **New `pre-2013-txt-row-completeness`** sharp edge documenting a row-completeness gap in the column-position TXT parser found on some pre-2013 filings (details incoming in a separate bug report) — the workaround is the same Entry Total reconciliation technique

No code changes — this PR only touches the two `edgar/ai/skills/holdings/*.yaml` files.

## Testing

Ran the actual test suite (`uv` env, since `hatch` wasn't already installed — same base dependency set, just a different resolver) against both test files that exercise these skill YAML files:

```
$ python -m pytest -m fast tests/test_ai_surface_verification.py tests/test_ai_skill_export.py
tests/test_ai_surface_verification.py ..........................  (26 passed)
tests/test_ai_skill_export.py ...........................         (27 passed)
============================= 53 passed in 1.88s ==============================
```

That covers YAML validity, required fields (`id`/`name`/`version`, `id` prefix), `skill.yaml` id uniqueness, the skills API (`list_skills`/`get_skill`), and the export/packaging path that bundles `sharp-edges.yaml` and `skill.yaml` into the exported skill package — all green on this branch.

Also checked, outside the test suite:
- Both `detection_pattern` regexes compile cleanly
- Every `sharp_edges` entry (including the new one) has `id`/`summary`/`severity`/`solution`, matching the existing convention
- `holdings/skill.yaml`'s token-budget check (`constitution.yaml`, 800-token cap): edited file estimates to ~518 tokens, comfortably under

## Related

Filed as #1072 — the pre-2013 TXT parser row-completeness gap that the new sharp edge documents.

### DIFF
--- a/edgar/ai/skills/holdings/sharp-edges.yaml
+++ b/edgar/ai/skills/holdings/sharp-edges.yaml
@@ -12,19 +12,40 @@ sharp_edges:
     detection_pattern: 'for.*in.*get_filings.*13F.*:.*\.obj\(\)'
 
   - id: value-in-thousands
-    summary: Holdings Value column is in $1000s, not dollars
+    summary: "Fixed in 5.28.2 (GH #749): Value is now normalized to whole dollars automatically -- do not multiply by 1000"
     severity: high
     solution: |
       holdings = thirteenf.holdings
-      total_value = holdings['Value'].sum() * 1000  # Convert to dollars
-    detection_pattern: "holdings\\['Value'\\]\\.sum\\(\\)(?!\\s*\\*\\s*1000)"
+      total_value = holdings['Value'].sum()  # Already in dollars, no conversion needed
+
+      # Before 5.28.2, the SEC's 13F XML schema change around Q4 2022 meant Value
+      # was in $1000s for report periods <= 2022-09-30 and dollars for later
+      # periods, and edgartools didn't normalize the difference -- so summing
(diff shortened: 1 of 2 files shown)


dgunning (approved):
Good catch. I have reviewed it as safe and necessary.

Appreciate the effort.

Dwight

PR #1073, opened 2026-08-21
fix(documents): read bare TextNodes as section headers on pre-2002 filings

dgunning:
Closes the last two parity gaps that kept the legacy `ChunkedDocument` fallback load-bearing in `company_reports/`.

## The bug

Filings from before roughly 2002 are preformatted text wrapped in minimal HTML. They parse to `ContainerNode > TextNode` with **zero** `HeadingNode`s and **zero** `ParagraphNode`s — and every header strategy in `_find_section_headers` draws its candidates from headings (S1), section nodes (S2), bold paragraphs (S3/3b), table cells (S4) or plain paragraphs (S5).

So on these documents there was no candidate source at all. Section detection returned nothing however well the patterns matched, and `TenK.items` / `TwentyF.items` came back empty unless the deprecated legacy parser rescued them.

## The fix

Strategy 5c reads bare `TextNode`s, gated on `_item_structure_found` being False so it cannot affect filings that already detect structure.

It takes each node's **first line** as the header. In a preformatted filing one node carries the heading *and* the body that follows it, so the untrimmed text is ~1000 characters of prose that `_looks_like_section_header` rejects on its 300-char cap. A title wrapped across two lines is truncated (`"...AND RESULTS"` without `"OF OPERATIONS"`), which costs nothing — the pattern match is anchored at the start and the section's title comes from the pattern table, not the matched text.

## Measured over all 121 corpus fixtures (tracked + parity_gate + era)

- **`.items` is identical with and without the `ChunkedDocument` fallback on 121/121, up from 118/121.** The `company_reports` legacy dependency is now behaviourally dead everywhere in the corpus, which is what `edgartools-3dp` needed.
- Two filings recover `Item 7` without the legacy parser: `0000927356-01-000369` (2001 10-K) and `0000928385-01-500187` (2001 20-F). Both `BASELINE_GAPS` entries are removed in this commit, as `test_closed_gaps_are_recorded` requires.
- One filing gains a genuine item: `0001012118-01-000019`, a 2001 10-Q, now yields `Part I, Item 2`. Legacy finds *nothing* on that filing, so the new parser beats it there. Verified against the filing text — the body carries `Item 2. Management's Discussion and Analysis` after a table close, and the earlier hit is the TOC entry.
- No item gained or lost anywhere else; no fixture starts erroring.

## Verification

- `tests/test_section_parity_ratchet.py` — 6 passed (the `test_closed_gaps_are_recorded` half was red before this change, by design)
- 109 section-detection tests pass across 8 files
- `hatch run test-fast` — 5367 passed, 4 skipped, 0 failures (up 4 = the new regression tests)
- Control run: with the fix reverted, both new item assertions fail; restored, they pass

**A check the existing tooling could not make.** `parity_benchmark` reports only `legacy_only` — what legacy finds and the new parser misses — so a change that *invents* items is invisible to it. "No new gaps" is not "no regressions". I captured `.items` across all 121 fixtures before and after and diffed both directions; that is how the gained item was found and then verified rather than assumed.

The new regression test anchors on the **tracked** `parity_gate` fixture so it runs in CI — its 10-K twin lives in the gitignored `text_boundary_corpus` and would skip silently there — and pins the zero-headings/zero-paragraphs precondition so a future parser change cannot leave it passing for a different reason.

## Not in scope

`("8-K", "0001104659-03-004925"): ["9"]` stays in `BASELINE_GAPS`. It blocks nothing: with `CurrentReport`'s Strategy 2 suppressed, all 18 8-K fixtures return identical `.items`, so that chunked union is dead weight rather than a fallback.

Separately noted and left alone: `find_toc_boundaries` searches for a literal lowercase `'</table>'`, and these filings carry 0 lowercase against 33 uppercase closes, so it falls through to a 50KB *estimate* of the TOC region. Real latent bug, unrelated to this fix, and a naive case-insensitive fix would regress the 10-K by stretching the region over the real heading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Filings from before about 2002 returned no items at all from the modern parser.** Those filings are preformatted text wrapped in minimal HTML, so they parse to a container-and-text tree with no headings and no paragraphs — and every header strategy in the section extractor drew its candidates from headings, section nodes, bold paragraphs or table cells. There was no candidate source at all on such a document, so detection returned nothing however well the patterns matched, and `TenK.items` or `TwentyF.items` came back empty unless the deprecated legacy parser rescued them. Bare text nodes are now read as headers, using each node's first line, since one node carries both the heading and the body that follows it. A 2001 10-K and a 2001 20-F recover their Item 7; across a 121-filing corpus this was the last remaining difference between the modern parser and the legacy one on these forms.
+
 ## [5.51.0] - 2026-08-19
 
 ### Performance
--- a/edgar/documents/extractors/pattern_section_extractor.py
+++ b/edgar/documents/extractors/pattern_section_extractor.py
@@ -755,6 +755,45 @@ class SectionExtractor:
                         headers.append((node, text.strip(), position))
                         existing_positions.add(position)
 
+        # Strategy 5c: bare TextNode headers, for filings with no block structure
+        # at all.
+        #
+        # Pre-2002 filings are preformatted text wrapped in minimal HTML, and they
+        # parse to ContainerNode > TextNode with *zero* HeadingNodes and *zero*
+        # ParagraphNodes. Every strategy above draws its candidates from headings,
+        # sections, bold paragraphs or table cells, so on those documents the
+        # header list is not merely short — there is no candidate source at all,
+        # and section detection returns nothing however good the patterns are.
+        # That is why these filings fell through to the legacy ChunkedDocument
+        # fallback in the report classes (edgartools-3dp).
+        #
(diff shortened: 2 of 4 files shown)


dgunning:
Follow-up: CI caught a regression the local verification missed.

The first commit applied Strategy 5c to every form. `test_issue_l6cl_8k_items_missing` failed on `regression (3.12)` — Strategy 5c can only find items that *start* a TextNode, so it may return a **partial** header set, and `CurrentReport.__getitem__` stops at whatever the new parser returns rather than falling through to its own text-based extraction. On GMAC `0001047469-05-006981` that meant asking for Item 4.02 also returned Item 8.01's text.

For annual forms a partial set still beats nothing; for current reports it is worse than nothing, since they already have era-appropriate text extraction. `895bf033` therefore excludes 8-K/6-K from 5c.

The PR body above describes the pre-regression verification. Corrected status: `hatch run test-regression` is green on `895bf033` locally — 2560 passed, 4 xfailed, 0 reruns.

Issue #462, opened 2025-10-15
[DATA] Incorrect items parsed from 8-K


Orgron:
## Data Quality Issue Details

**Issue Type:**
- [x] Incorrect financial values (wrong numbers)
- [ ] Missing financial data (expected data not present)
- [ ] Calculation errors (formulas producing wrong results)
- [ ] Data inconsistency (different values for same metric)
- [ ] Historical data problems (changes over time)

## Environment
**EdgarTools Version:** 4.19.1
**Python Version:** 3.12.10  
**Operating System:** Windows 10

## Financial Data Details
**Company/Ticker:** CIK 864509
**Form Type:** 8-K
**Filing Date/Period:** 1999-10-13
**Statement Type:** 8-K

**Specific Metric/Concept:**
- Financial line item:  "items"
- XBRL concept name: -

## Data Issue
**Expected Value:** 
- items: 7
- Source: [SEC filing](https://www.sec.gov/Archives/edgar/data/864509/0000906197-99-000155.txt)
- Page/section reference: 1

**Actual Value from EdgarTools:**
- items: 3
- How obtained: `edgar.Company(864509).get_filings(form="8-K", trigger_full_load=True, amendments=False, filing_date='1999-10-13')[0].items`

**Code to reproduce:**
```python
import edgar

print('Incorrect value:',edgar.Company(864509).get_filings(form="8-K", trigger_full_load=True, amendments=False, filing_date='1999-10-13')[0].items)
```

## Cross-Verification
**Have you verified this issue with:**
- [ ] Multiple time periods for same company
- [ ] Multiple companies with same issue
- [x] Direct SEC filing comparison
- [ ] Other financial data sources

**Affects multiple periods/companies?**
- Companies tested: just one
- Time periods tested: just one
- Pattern observed: none

## Expected Behavior
**What should happen:**
The `items` field should be filled with the string "7"

**Data validation rules:**
None

## Additional Context

**Impact Assessment:**
- [ ] Minor (affects specific edge case)
- [ ] Moderate (affects common use cases)  
- [ ] Major (affects core financial calculations)
- [x] Critical (produces completely wrong results)

---
*Data quality issues are high priority and will be verified against official SEC filings. Accuracy is fundamental to EdgarTools.*

dgunning:
@Orgron Thank you for the extremely detailed bug report! This is excellent documentation.

[...]

PR #1098, opened 2026-08-23
refactor(company_reports): delete the item-lookup fallbacks to edgar.files (3dp Group B)

dgunning:
`edgartools-3dp` Group B. Six first-hit fallbacks, plus the `id_parse_document` layer stacked beneath two of them.

Each tried the modern parser and, only on a miss, read the filing again through `edgar.files` and returned:

| File | Method |
|---|---|
| `ten_k.py` | `__getitem__`, `get_item_with_part` |
| `ten_q.py` | `__getitem__`, `get_item_with_part` |
| `twenty_f.py` | `__getitem__` |
| `current_report.py` | `__getitem__` (Strategy 2) |

## Measured, not assumed

The bead cites a 115-filing measurement from PR #1074. **That corpus is not in the tree** — `parser_corpus/manifest.json` holds 54, all 10-K and 10-Q — so rather than inherit the claim I re-measured, running every lookup twice: once normally, once against a throwaway subclass whose `edgar.files` paths are unavailable.

| Coverage | Lookups | Fallback contributed |
|---|---|---|
| 10-K/10-Q `__getitem__` (55 local fixtures) | 892 | 0 |
| 10-K/10-Q part-qualified | 690 | 0 |
| 20-F (network, 12 filings) | 96 | 0 |
| 8-K (network, 12 filings) | 72 | 0 |
| 8-K 2001 / 2004 / 2008 | 360 | 0 |

The legacy paths were **reached 86 times** across these and returned nothing on every one.

**The result is not vacuous.** `same` would also count "both returned `None`", so I measured the content-bearing share separately: **610 of 619 `__getitem__` lookups (98%) returned real content, and all 610 were byte-identical without the fallback.**

Two of the six sites — `twenty_f` and `current_report` — have **no local fixture coverage at all**, so those numbers come from filings fetched over the network.

### The one difference found anywhere was whitespace

8-K `0000893220-08-001974` Item 5.02: 404 characters with the fallback against 395 without. Identical words, different line wrapping — `"".join(a.split()) == "".join(b.split())` holds exactly.

Old 8-Ks were the real risk case, because `current_report`'s Strategy 2 sat in front of the text-based path that carries pre-HTML SGML filings (GH #462). **That Strategy 3 is untouched** and now runs as Strategy 2.

## A guard test changed meaning

`TestGetitemFallbackIsStillWired` pinned this fallback as *live*, including that "TenK dereferences it unguarded where TwentyF returns None, the difference 07lk.3 has to handle at both call sites."

This handles it by deletion: with nothing left to dereference, every report type returns `None` for an item it does not have. **`TenK` previously raised `TypeError` there** — the one behaviour change in this PR.

The class is rewritten and renamed rather than deleted. Its record of when each of the fifteen legacy-only lookups closed is precisely what made removing this measurable rather than hopeful.

## Deliberately out of scope

- **`current_report.py:718` `list_items()`** — Group C. A three-way *union* that backfills items the modern parser missed, so deleting it can silently lose items in a way none of these can. The bead says do not start there.
- **`_base.py:217` `chunked_document`** — Group D, the user-facing break belonging to `07lk.22`. Still deprecated-but-available; the `ChunkedDocument` imports stay for these two.

## Verification

Full fast suite: **5511 passed, 0 failed.**


dgunning:
## Added: 20-F and 8-K fixtures, so this PR's evidence runs in CI

The original PR body noted that the 20-F and 8-K numbers came from filings fetched over the network, which no CI job can repeat — there were no fixtures for either form in the tree. Commit `513f17f0` closes that.

**12 fixtures, 2.42MB total** (tree 277MB → 280MB):

| Form | Count | Sizes | Coverage |
|---|---|---|---|
| 20-F | 3 | 0.19 / 0.57 / 1.45 MB | three different filer agents |
| 8-K | 9 | 3.7–59.5 KB | 2001, 2004, 2008, 2025 |

Deliberately small: a mega-cap 20-F runs 6–24MB (Shell 9.8, Alibaba 11.2, Rio Tinto 23.6), so these are ordinary foreign private issuers — cheaper, and more representative of the form than Shell is. The 8-K era spread is chosen because `CurrentReport` ran the deleted fallback *ahead* of the text strategy for pre-HTML SGML filings (GH #462), so the oldest filings were the ones most likely to have needed it.

### Two mistakes in the test, both caught by checking

Recording these because the same trap will recur in Group C.

**1. The obvious test is tautological.** My first version parsed each fixture twice — once with `_chunked_document` nulled — and compared. After the deletion there is no branch left that reads it, so the two runs are identical *by construction* and the test can never fail. That comparison was migration evidence, meaningful only *before* the fallback was removed.

**2. The replacement guard didn't work either.** `_forbidding_legacy` raises if `_chunked_document` is read. I reintroduced the deleted branch into `twenty_f.py` expecting failures and got **17 passes** — because every pinned item is answered by the modern parser, so the fallback branch is never reached. The guard only bites on a lookup that *misses*, and that test existed for 8-K but not 20-F.

With `test_a_missing_twenty_f_item_returns_none_without_reading_legacy` added, reintroducing the fallback fails exactly those three tests and removing it makes them pass — verified in both directions.

Lengths are asserted as **floors, not exact counts**: exact item lengths move whenever table rendering changes (#1095 and #1097 each moved a batch), and this file is about which parser answers, not how wide a table renders.

Full fast suite: **5531 passed, 0 failed.**

Issue #931, opened 2026-07-31
bs4 → lxml migration program: EdgarTools 6.0 ships bs4-free


dgunning:
## Why

EdgarTools parsed XML and HTML with BeautifulSoup in 32 files. Where we migrated to lxml the wins were large and measured: the 13F infotable parser got **8–20x faster**, and the EX-FILING FEES exhibit parser ~15x with byte-identical output. 6.0 ships **bs4-free**, and `beautifulsoup4` comes off the dependency list.

The full public inventory — every file, sized S/M/L, in migration order — is at [`engineering/analysis/bs4-migration-inventory.md`](https://github.com/dgunning/edgartools/blob/main/engineering/analysis/bs4-migration-inventory.md).

## Status: the ports are done; the dependency drop is a consequence, not a project

1. ~~`edgar/xmltools.py` first — 12 files depend on it.~~ **Done**, as a dual-backend adapter that accepts either a bs4 `Tag` or an lxml element, so its ~350 call sites never had to change at once.
2. ~~Hot-path parsers (ownership, funds, offerings).~~ **Done.** The nine XML parsers shipped in **5.51.0** at 2.7–9.0x (table in the comments).
3. ~~Long tail of per-file migrations.~~ **Done.** The HTML half landed across 5.52.0–5.55.0, including `markdown.py` and `funds/data.py`, the pair the inventory flagged as large and high-risk.
4. **Drop the `beautifulsoup4` dependency in the 6.0 breaking window.** ← all that is left

Measured on `main` at 5.56.0, exactly **five** real `from bs4` imports survive in the library: four inside the legacy `edgar.files` package, which #930 deletes, and the bs4 half of the `xmltools` dual-backend adapter, whose own module docstring already schedules it for this window now that every dependent is migrated. So step 4 is the removal of one package plus one adapter half — there is no third body of work between the finished ports and the dependency leaving `pyproject.toml`.

Two open follow-ups remain, and neither is a port or a blocker: folding the five per-file text helpers into shared `html_utils`, and one schema-element parsing gap the migration *surfaced* rather than caused (MA-I `isEngagedInOtherBusiness` / `isDrpRemovable`, never read by either parser).

## What the migration found

[...]

PR #1107, opened 2026-08-23
refactor(headers): port edgar/headers.py to lxml.html (GH #1101)

### LINKED ISSUE #1101: 6.0 bs4→lxml: port edgar/headers.py to lxml.html (calibration port)
Part of the bs4 → lxml migration program (#931). **Read the [porting guide](https://github.com/dgunning/edgartools/issues/931#issuecomment-5387920956) first** — it carries the six gotchas that make these ports risky, and they are not optional reading.

**This is the calibration port** — the first of the HTML half, and the one that tells us whether the effort model for the remaining files is right. Worth taking if you want the one with the most attention on it.

`edgar/headers.py` parses SGML filing-header text and `<pre>` blocks as HTML. Current bs4 usage:

- the import at line 6 (`from bs4 import BeautifulSoup, Comment, Tag`)
- `BeautifulSoup(header_text, 'html.parser')` at ~line 264
- `BeautifulSoup(pre_tag.text)` at ~line 483 — note that this one passes **no explicit parser argument** today

### Two file-specific traps

- **This file handles bs4 `Comment` nodes explicitly.** With `remove_comments=True` comments vanish at parse time and libxml2 *preserves their tails*, whereas removing them manually after parsing walks straight into gotcha 2 and loses those tails.
- **The `BeautifulSoup(pre_tag.text)` call at ~:483 parses arbitrary `<pre>` content**, which makes it the likeliest site in this phase for gotcha 5 — a fragment root, or an empty input that now raises where bs4 returned an empty soup. Cover both in the characterization tests.

### Done looks like

Behaviour identical, characterization tests written *before* the port where coverage is thin, and a measured before/after on a real workload (e.g. parsing N cached filing headers). Plan ~2–3x on the parse step.

After it lands, please include your actual effort and measured speedup in the PR — we re-estimate the remaining HTML files from it.

ryanduguid:
Closes #1101. Part of the bs4 → lxml migration program (#931).

## What changed

`IndexHeaders.load()` no longer uses BeautifulSoup. The SEC-HEADER payload is extracted from the HTML comment by a new `IndexHeaders._extract_sec_header_comment()` helper built on `lxml.html.HTMLParser(remove_blank_text=False, remove_comments=False, recover=True)`, the same pattern as `edgar/documents/parser.py:create_lxml_parser` per the [porting guide](https://github.com/dgunning/edgartools/issues/931#issuecomment-5387920956).

- `from bs4 import ...` is gone; `edgar/headers.py` now imports only lxml.
- The two dead bs4-typed helpers (`_extract_documents_from_pre`, `_extract_comment_text`) are removed. Neither was referenced except from a commented-out line; keeping either would have kept the bs4 import alive.
- Empty / whitespace-only / comment-less input still raises `IndexError`, matching the behaviour pinned before the port. Under lxml an empty document raises `etree.ParserError`, so that path is mapped explicitly.

## Characterisation first

Per the guide, the pre-port output was pinned before the swap: `tests/test_headers_characterization.py` (commit 48f5b24) holds field-by-field assertions over four real fixtures plus the edge-case contracts:

- 23andMe 8-K (`0001193125-24-152391`) and Consumers Energy 144 (`0001971857-23-000246`) pinned for every leaf including nested `filing_values` on the subject company
- bare-SGML Form 4 fixture exercising the comment-missing path
- empty input, whitespace-only input, and HTML-without-comment all pinned to `IndexError`

All 15 tests (9 characterisation + 6 existing in `test_headers.py`) pass unchanged after the port.

## Measured before/after

Fixture workload: the five real index-header pages in `data/headers/`, best of repeated runs, Python 3.12, Windows.

| scope | bs4 | lxml.html | change |
|---|---|---|---|
| parse + extract step only | 0.109 ms | 0.060 ms | **1.84x** |
| full `IndexHeaders.load()` end-to-end | 0.117 ms | 0.066 ms | **1.78x** |

In line with the guide's ~2-3x projection for the HTML half. Side benefit: the spurious `MarkupResemblesLocatorWarning` bs4 emitted on small headers is gone.

## Guide gotchas encountered

Two of the six bit during the port, both handled: fragment auto-detection raising on empty input where bs4 returned an empty soup (gotcha 5, mapped to the pinned IndexError), and comment nodes' callable `.tag` (gotcha 6, compared directly against `etree.Comment`).

## Verification

- `pytest tests/test_headers_characterization.py tests/test_headers.py`: 15 passed
- `ruff check edgar/headers.py`: clean (ruff 0.12.8 per .pre-commit-config.yaml)
- No new bs4 imports; net one fewer file importing bs4

### DIFF
--- a/edgar/headers.py
+++ b/edgar/headers.py
@@ -3,7 +3,8 @@ from typing import Any, Dict, List, Optional
 
 import orjson as json
 import pandas as pd
-from bs4 import BeautifulSoup, Comment, Tag
+from lxml import etree
+from lxml import html as lxml_html
 from pydantic import BaseModel
 from rich import box
 from rich.console import Group
@@ -14,6 +15,7 @@ from rich.text import Text
 from edgar._party import Address, get_addresses_as_columns
 from edgar.config import SEC_BASE_URL
 from edgar.display.formatting import display_size
+from edgar.documents.utils.html_utils import remove_xml_declaration
 from edgar.httprequests import download_file
 from edgar.reference import describe_form
 from edgar.richtools import repr_rich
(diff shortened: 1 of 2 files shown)


ryanduguid:
Closing this one - upstream #1105 landed the same port while this was in flight: lxml.html with recover mode, the pinned IndexError contract on empty/comment-less input, and characterisation coverage written against bs4 first. That implementation covers everything this branch did (and uses golden files for the fixture dump, which goes further than my field assertions), so there is nothing left worth merging here. Thanks @dgunning - enjoyed working through the porting guide, the six gotchas were exactly the right briefing.

Issue #1104, opened 2026-08-23
6.0 bs4→lxml: port edgar/xbrl/notes.py to lxml.html


dgunning:
Part of the bs4 → lxml migration program (#931). **Read the [porting guide](https://github.com/dgunning/edgartools/issues/931#issuecomment-5387920956) first** — it carries the six gotchas that make these ports risky, and gotchas 1 and 2 are the ones that will get you here.

**This is the highest-risk of the four — take it if you want the interesting one, not the easy one.**

`edgar/xbrl/notes.py` extracts note text and tables, and its output is **RAG-facing text** that people feed to models. Whitespace semantics differ between bs4's `get_text()` and lxml's `text_content()`, and this codepath family has a history of delete-versus-collapse whitespace bugs.

So the acceptance bar here is higher than elsewhere: **pin the current emitted text on a corpus of real notes before switching, and parity-test it exactly** — not "close enough", byte-identical.

Current bs4 usage:

- local imports `from bs4 import BeautifulSoup, NavigableString` at ~lines 789 and 887
- `BeautifulSoup(html, 'html.parser')` at ~line 791
- the html-table-to-aligned-plain-text conversion at ~line 841

### On the expected payoff

This file is traversal-dominated rather than parse-dominated, so expect the **smaller end** of the speedup range. The reason to do it is the dependency, not the milliseconds — worth knowing before you measure and feel underwhelmed.

### Done looks like

Byte-identical emitted text over a real notes corpus, characterization tests written *before* the port, and a measured before/after in the PR.

dgunning:
Done in #1111, merged a few hours after this was filed — **477 ms → 80 ms over 16 real note TextBlocks, ~6x**, both sides timed on the same machine by stashing the port. The smaller multiple than the other three is exactly what this issue predicted: the path is traversal-dominated.

Apologies for the churn: this was labelled `help wanted` and then picked up in-house the same evening, which is the wrong signal to send.

The correctness half was the real work, as expected for RAG-facing text. Four bs4 behaviours turned out to have no lxml equivalent and needed deliberate translation rather than substitution — starting with `get_text(separator=' ', strip=True)`, which is **not** `text_content()`: the latter concatenates without separators, so a naive swap silently glues words together at element boundaries. That is gotcha 1 from the porting guide showing up in production code.

PR #1118, opened 2026-08-24
perf(attachments): parse filing index pages with lxml instead of bs4

### LINKED ISSUE #1102: 6.0 bs4→lxml: port edgar/attachments.py to lxml.html
Part of the bs4 → lxml migration program (#931). **Read the [porting guide](https://github.com/dgunning/edgartools/issues/931#issuecomment-5387920956) first** — it carries the six gotchas that make these ports risky, and they are not optional reading.

`edgar/attachments.py` parses attachment index-page HTML, and it is a **hot path**: it runs on every filing's attachment listing.

Current bs4 usage:

- the import at line 23
- `def load(cls, soup: BeautifulSoup)` at ~line 951 — this puts a bs4 type in an internal signature, so check and adjust every caller
- a `soup` parameter at ~line 1073
- `BeautifulSoup(response.text, 'html.parser')` at ~line 1221

### Done looks like

Behaviour identical, characterization tests written *before* the port where coverage is thin, and a measured before/after on a real workload. Plan ~2–3x on the parse step.

Because this is a hot path, **measure memory as well as time** — the XML half saw several-x memory reductions alongside the speedups, and on a path that runs for every filing that is worth reporting separately in the PR.

ryanduguid:
Closes #1102. Part of the bs4 → lxml migration program (#931).

## What changed

- **New `edgar/_index_parsing.py`**: the house parser settings (`recover=True`, `remove_blank_text=False`, XML declaration stripped via `remove_xml_declaration`) plus small find/text helpers that keep bs4 semantics:
  - class matching is on **whole tokens** (XPath `concat(' ', @class, ' ')` trick), so `class="infoHead"` does not match `class_='info'` — the string-vs-list API false friend from gotcha 6
  - `element_text()` skips comment payloads exactly like bs4's `.text`, and preserves tails
  - empty/whitespace input raises `IndexError`, matching what the old code produced downstream on an empty soup (gotcha 5)
- `Attachments.load(soup)` and `FilingHomepage(soup=...)` now take a parsed lxml tree; `FilingHomepage.load(url)` parses with the house parser
- identInfo `<br>` newline handling sets tail text instead of `replace_with`, which lxml does not support
- The sc13g multi-filer regression test's fixture builder moved to the same helper (caller adjustment per the issue)

## Proof nothing changed

Characterization tests written **before** the port (`4a9d3dc`), pinning document tables, XBRL datafiles tables, filer extraction and filing dates on three real index pages:

| fixture | covers |
|---|---|
| apple 10-Q index | iXBRL primary doc, GRAPHIC rows, complete-submission row, 6-file datafiles table |
| fund 485 (8-K) index | exhibits, graphics, 5-file datafiles table |
| T. Rowe Price DEF 14A | single-form proxy, no datafiles, **66 filer divs** |

All green unchanged after the swap, plus the sc13g multi-filer regression suite and `test_filing.py`.

Full fast suite: 5578 passed, 13 failed — the same-by-name set that fails on pristine `main` on this Windows box (network-marked filing-text baselines + two file-lock tests), verified before any edits.

## Measured

Full caller path (parse + Attachments + get_filers + get_filing_dates), median of 200 runs per page:

| page | before (bs4) | after (lxml) | speedup | peak allocation |
|---|---|---|---|---|
| apple 10-Q index | 16.9 ms | 1.6 ms | **10.7x** | 22x less |
| fund 485 index | 20.1 ms | 1.9 ms | **10.4x** | 22x less |
| troweprice DEF 14A (66 filers) | 657.4 ms | 29.6 ms | **22.2x** | 30x less |

Memory measured with tracemalloc peak alongside wall time, since this path runs for every filing.

### DIFF
--- a/edgar/_index_parsing.py
+++ b/edgar/_index_parsing.py
@@ -1,0 +1,174 @@
+"""
+lxml.html helpers for parsing SEC filing index pages.
+
+Part of the bs4 -> lxml migration (#931, #1102). These helpers adapt lxml's
+API to the find/text idioms the attachment index parser was written against,
+while preserving bs4-compatible behaviour on class attributes and whitespace.
+
+The house rules from edgar/documents/parser.py apply:
+
+- parse with recover=True so broken SEC HTML still yields a tree
+- never remove blank text: a whitespace-only node between two tags is a
+  word boundary ("Yes ☒", "three reportable")
+- strip any XML declaration first: lxml raises ValueError on a str that
+  starts with one (gotcha 4 in the #931 porting guide)
+- empty or unparseable input raises ParserError where bs4 returned an empty
+  soup; callers here translate that to IndexError, matching what the old
+  bs4-based code produced on those inputs
+"""
+
+from typing import List, Optional
+
+import lxml.html
+from lxml.etree import ParserError
+
+from edgar.documents.utils.html_utils import create_lxml_parser, remove_xml_declaration
+
+__all__ = [
+    'parse_index_html',
+    'find_all',
+    'find_one',
+    'element_text',
+    'class_tokens',
+]
+
+
+def parse_index_html(html: str) -> lxml.html.HtmlElement:
+    """
+    Parse an SEC filing index page into an lxml tree.
+
+    Uses the house parser settings (recover=True, remove_blank_text=False)
+    so that malformed filings still parse and whitespace-only text nodes
+    survive as word boundaries.
+    """
+    if html is None:
+        raise IndexError("No HTML content to parse")
+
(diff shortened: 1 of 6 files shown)


ryanduguid:
Closing: upstream landed the same port in #1109 (merged 2026-08-23, ~8.9x) plus #1108's get_filers fix while this was in flight. My characterization fixtures (apple 10-Q / fund 485 / troweprice DEF 14A with 66 filer divs) and the tracemalloc peak-alloc numbers are there if useful for #1104 or any remaining file — happy to re-point them. Good luck with 6.0.

PR #1119, opened 2026-08-24
fix(xbrl): suppress low-confidence fuzzy matches onto total concepts

### LINKED ISSUE #914: Line-item tags mapped to balance-sheet total concepts — `standard_concept == 'Assets'` returns 3 rows on a bank filing
Line-item tags are mapped to balance-sheet *total* concepts, so a standardized statement returns several rows claiming to be the total. Affects `standard_concept` on any statement dataframe.

Version 5.43.0.

## Repro

```python
from edgar import Company

df = Company('CFG').get_financials().balance_sheet().to_dataframe()
df[df['standard_concept'] == 'Assets'][['concept', 'label', '2025-12-31']]
```

Citizens Financial Group, FY2025 — three rows:

| concept | label | 2025-12-31 |
| --- | --- | ---: |
| `us-gaap_DebtSecuritiesHeldToMaturityExcludingAccruedInterestAfterAllowanceForCreditLoss` | Debt securities held to maturity | 7,933,000,000 |
| `us-gaap_BankOwnedLifeInsurance` | Bank-owned life insurance | 3,441,000,000 |
| `us-gaap_Assets` | TOTAL ASSETS | 226,351,000,000 |

Selecting on the standardized concept and taking the first row reports total assets as $7.9B instead of $226.4B.

No network needed:

```python
from edgar.xbrl.standardization.core import MappingStore
MappingStore().get_standard_concept('us-gaap_BankOwnedLifeInsurance')   # 'Total Assets'
MappingStore().get_standard_concept('ifrs-full_IntangibleAssetsAndGoodwill')  # 'Total Non-Current Assets'
```

`edgar/xbrl/standardization/reverse_index.py`. 53 index entries sit in `section: 'Totals'` with `is_total: True`. Most are line items:

```
Total Assets            Assets, AssetsNet, ifrs:Assets
                        + AccruedInvestmentIncomeReceivable, BankOwnedLifeInsurance,
                          CashCashEquivalentsAndFederalFundsSold, FederalFundsSold,
                          FederalHomeLoanBankStock, InvestmentOwnedAtFairValue,
                          DebtSecuritiesHeldToMaturityExcludingAccruedInterest...

Total Liabilities       Liabilities, ifrs:Liabilities
                        + AdvancesFromFederalHomeLoanBanks, DueToRelatedPartiesCurrentAndNoncurrent,
                          ManagementFeePayable, OtherBorrowings, SecuredDebt,
                          SecuritiesSoldUnderAgreementsToRepurchase, SubordinatedDebt,
                          UnearnedPremiums, WarrantLiability,
                          SharesSubjectToMandatoryRedemptionSettlementTermsFairValueOfShares

Total Current Liabilities   LiabilitiesCurrent, CurrentLiabilities
                        + CurrentBorrowingsAndCurrentPortionOfNoncurrentBorrowings,
                          CurrentDerivativeFinancialLiabilities, CurrentPortionOfLongtermBorrowings,
                          CurrentProvisions, CurrentTaxLiabilitiesCurrent,
                          LiabilitiesIncludedInDisposalGroupsClassifiedAsHeldForSale,
                          OtherCurrentFinancialLiabilities, TemporaryEquityLiquidationPreference,
                          TradeAndOtherCurrentPayablesToRelatedParties

Total Current Assets    AssetsCurrent, CurrentAssets
                        + CurrentDerivativeFinancialAssets, CurrentPrepaymentsAndOtherCurrentAssets,
                          CurrentTaxAssetsCurrent, CurrentTradeReceivables, OtherCurrentReceivables,
                          CurrentAssetsOtherThanAssetsOrDisposalGroups...

Total Non-Current Assets    AssetsNoncurrent, NoncurrentAssets, ifrs:NoncurrentAssets
                        + IntangibleAssetsAndGoodwill, NoncurrentDerivativeFinancialAssets
```

The concentration is in bank and insurer line items (federal funds sold, FHLB stock, BOLI, unearned premiums, subordinated debt), which is why it doesn't show on a typical industrial filer. `us-gaap_Goodwill` maps correctly to `Goodwill`; the mis-mapped set is specific.

Separately, `_normalize_tag` (reverse_index.py:188) strips `ifrs-full_` / `ifrs-full:` before lookup, so IFRS tags resolve against the GAAP index by bare name. On VEON's 20-F, `ifrs-full_IntangibleAssetsAndGoodwill` and `ifrs-full_CurrentProvisions` are labeled Total Non-Current Assets and Total Current Liabilities. VEON has 24 of 111 rows standardized, Klarna 16 of 81, so IFRS coverage is thin to begin with, the wrong mappings are the part that's actively misleading rather than just missing.

Random sample of 60 tickers from `get_cik_tickers()`, checking `Assets == Liabilities + Equity` per rendered period. The identity holds everywhere (0 mismatches over 98 periods, plus 49 large caps) because it reads raw concepts, the defect is only in the standardization layer.

Is `gaap_taxonomy_mapping.csv` the right place to fix this, or the index generation? The `section: 'Totals'` bucket looks like it's doing double duty, "this tag belongs to the totals region of the balance sheet" vs "this tag *is* the total", and `is_total: True` on a component is what makes it selectable as one. If the intended fix is in the source CSV I can't regenerate the index from the repo, since `data/xbrl-mappings/` isn't checked in.

Happy to send a PR once you've said which layer it belongs in. A cheap guard in the meantime would be refusing `is_total: True` for a tag whose `standard_tags` is a `*Total` concept it isn't the canonical tag for, but that's a patch on the output.

ryanduguid:
Fixes #914 (the GAAP half; tracked as `edgartools-gnx5`).

## The bug

On Citizens Financial Group's balance sheet, `standard_concept == 'Assets'` returned **three rows**:

| concept | label | 2025-12-31 |
|---|---|---:|
| `us-gaap_DebtSecuritiesHeldToMaturityExcludingAccruedInterestAfterAllowanceForCreditLoss` | Debt securities held to maturity | 7,933,000,000 |
| `us-gaap_BankOwnedLifeInsurance` | Bank-owned life insurance | 3,441,000,000 |
| `us-gaap_Assets` | TOTAL ASSETS | 226,351,000,000 |

Selecting the standardized total reported **$7.9B as total assets** instead of $226.4B.

## Root cause (per the #914 thread)

`gaap_mappings.json` carries 59 entries at confidence < 0.5 whose `standard_tags` are canonical totals (`Assets`, `Liabilities`, `CurrentAssetsTotal`, ...). The file is generated upstream (`edgar-storage StandardizedConcepts` v3.0.0) and hand edits would not survive regeneration, so the authority judgement belongs in the consumer.

## The change

`ReverseIndex.lookup` now suppresses a mapping when **confidence < 0.5 and every candidate concept is a total concept**. The floor comes from the distribution measured in the #914 thread: in the 65-filing sample, every colliding row sat at confidence 0.307–0.319 and nothing between 0.5 and 0.9 collided — the mid band is harmless because context disambiguation catches it before it surfaces (`CommonStocksIncludingAdditionalPaidInCapital` at exactly 0.5 resolves correctly to `CommonEquity`).

Scope on the current index: **59 of 2,924 entries** suppressed, all in the 0.307–0.499 band with exclusively-total candidates. Untouched: canonical mappings (`Assets`/`Liabilities`/`AssetsCurrent` at ≥ 0.9), `AssetsNet` at exactly the 0.5 floor, non-total low-confidence mappings, and entries with no confidence field (treated as no signal).

## Tests

New `tests/issues/regression/test_issue_914_total_false_matches.py` (written first, RED on `main` — 12 failing bug assertions, 3 passing guards):

- the 10 colliding tags from the issue report + follow-up sampling must not resolve to their total
- the IFRS half of #914: `ifrs-full_IntangibleAssetsAndGoodwill` no longer lands on `NonCurrentAssetsTotal` via its 0.314-confidence GAAP entry (a namespace-aware IFRS index remains upstream work)
- canonical tags still resolve; `AssetsNet` at exactly 0.5 still resolves; mid-band tag still resolves

`test_standardization.py` (33 tests) + `test_standardization_integration.py` green unchanged.

## Verification

Full fast suite on this branch: **5906 passed, 19 failed**. The 19 are the 13 known environmental failures on this Windows box (verified against clean `main` earlier: network-marked filing-text baselines + file locks) **plus 6 in upstream's own new characterization files** (`test_homepage_characterization.py`, `test_drs_characterization.py`), which I verified fail identically on pristine `3e910f8` — unrelated to this change, they came in with the #1110–#1115 lxml ports.

Offline acceptance (no network): the CFG three-row repro now yields exactly one row claiming `Assets` — the canonical one.

### DIFF
--- a/edgar/xbrl/standardization/reverse_index.py
+++ b/edgar/xbrl/standardization/reverse_index.py
@@ -54,6 +54,17 @@ def _is_total_concept(standard_concept: str) -> bool:
     return standard_concept in _TOTAL_CONCEPT_IDS or 'Total' in standard_concept
 
 
+def _is_low_confidence_total_mapping(xbrl_tag: str, entry: dict) -> bool:
+    """True for a non-canonical fuzzy mapping whose candidates are all totals."""
+    standard_tags = entry.get("standard_tags", [])
+    return (
+        entry.get("confidence", 1.0) < LOW_CONFIDENCE_TOTAL_FLOOR
+        and bool(standard_tags)
+        and xbrl_tag not in standard_tags
+        and all(_is_total_concept(tag) for tag in standard_tags)
+    )
+
+
 @dataclass
 class MappingResult:
     """Result of a reverse index lookup."""
@@ -148,6 +159,14 @@ class ReverseIndex:
 
         # Build the reverse index
         self._index: Dict[str, dict] = self._gaap_mappings
+
+        # Suppression depends only on the generated base mappings. In particular,
+        # a high-confidence industry override must not revive a fuzzy total match.
+        self._suppressed_total_mappings: Set[str] = {
+            tag
+            for tag, entry in self._index.items()
+            if isinstance(entry, dict) and _is_low_confidence_total_mapping(tag, entry)
+        }
 
         # Cache for normalized lookups (strips namespace prefixes)
         self._normalized_cache: Dict[str, str] = {}
@@ -287,42 +306,18 @@ class ReverseIndex:
             return None
 
         # Apply industry overrides if industry is known
-        if industry and isinstance(entry, dict):
+        if industry:
             entry = self._apply_industry_override(entry, industry)
+
+        if normalized in self._suppressed_total_mappings:
+            logger.debug("Suppressed low-confidence total mapping for %s", xbrl_tag)
+            return None
 
         # Extract data from entry
         standard_tags = entry.get("standard_tags", [])
         is_ambiguous = entry.get("ambiguous", False)
         deprecated = entry.get("deprecated")
         comment = entry.get("comment")
-
-        # Low-confidence fuzzy matches onto total concepts are not
-        # authoritative (GitHub #914): on a bank balance sheet,
-        # BankOwnedLifeInsurance resolved to 'Assets', so
-        # standard_concept == 'Assets' matched line items worth tens of
-        # billions alongside the real TOTAL ASSETS row. gaap_mappings.json
-        # is generated upstream and hand edits would not survive
-        # regeneration, so the authority judgement belongs here in the
-        # consumer: below the floor, a mapping whose candidates are all
-        # total concepts is suppressed outright. The confidence
-        # distribution measured in #914 is cleanly split - colliding rows
-        # all sat at 0.307-0.319, nothing between 0.5 and 0.9 collided -
-        # so the floor clears every reproducible case while the mid-band
-        # (harmless, context-disambiguated) and canonical mappings are
-        # untouched. Missing confidence means no signal: keep the mapping.
-        entry_confidence = entry.get('confidence', 1.0) if isinstance(entry, dict) else 1.0
-        if (
-            isinstance(entry, dict)
-            and entry_confidence < LOW_CONFIDENCE_TOTAL_FLOOR
-            and standard_tags
-            and all(_is_total_concept(tag) for tag in standard_tags)
-        ):
-            logger.debug(
-                "Suppressed low-confidence total mapping for %s "
-                "(confidence=%.3f, candidates=%s)",
-                xbrl_tag, entry_confidence, standard_tags
-            )
-            return None
 
         # Resolve display names: prefer embedded display_name, then display_names.json, then concept name
         display_names = []
@@ -457,7 +452,7 @@ class ReverseIndex:
             # Phase 4: Special case - if is_total is True (from context or entry), prefer "Total" concepts
             if is_total or entry_is_total:
                 for candidate in candidates:
-                    if 'total' in candidate.lower():
+                    if _is_total_concept(candidate):
                         logger.debug(
                             "Disambiguated %s to %s (is_total=True)",
                             xbrl_tag, candidate
--- a/edgar/xbrl/standardization/reverse_index.py
+++ b/edgar/xbrl/standardization/reverse_index.py
@@ -28,6 +28,31 @@ from .exclusions import should_exclude, EXCLUDED_TAGS
 
 logger = logging.getLogger(__name__)
 
+# Below this confidence a mapping whose candidates are all total concepts is
+# suppressed (GitHub #914). The colliding rows measured in #914 sit at
+# 0.307-0.319; nothing between 0.5 and 0.9 collided in a 65-filing sample.
+LOW_CONFIDENCE_TOTAL_FLOOR = 0.5
+
+# Concept ids that name an aggregate rather than a line item. A fuzzy match
+# onto one of these from a non-total tag is the #914 failure shape.
+_TOTAL_CONCEPT_IDS = {
+    'Assets',
+    'Liabilities',
+    'CommonEquity',
+    'AllEquityBalance',
+    'AllEquityBalanceIncludingMinorityInterest',
+    'LiabilitiesAndEquity',
+    'CurrentAssetsTotal',
+    'NonCurrentAssetsTotal',
+    'CurrentLiabilitiesTotal',
+    'NonCurrentLiabilitiesTotal',
+}
+
+
+def _is_total_concept(standard_concept: str) -> bool:
+    """True if the concept id names an aggregate (total) rather than a line item."""
+    return standard_concept in _TOTAL_CONCEPT_IDS or 'Total' in standard_concept
+
 
 @dataclass
 class MappingResult:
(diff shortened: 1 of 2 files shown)


dgunning:
## Code review — 7 findings (verified against the branch)

The two load-bearing findings mean the guard doesn't hold in either of its main scenarios as written.

### 1. 🔴 The guard suppresses the canonical `LiabilitiesAndEquity` mapping (`edgar/xbrl/standardization/reverse_index.py:314`)

`gaap_mappings.json`'s `LiabilitiesAndEquity` entry is a placeholder **self-match** (`{'standard_tags': ['LiabilitiesAndEquity'], 'confidence': 0.4, 'company_count': 0}`). 0.4 < `LOW_CONFIDENCE_TOTAL_FLOOR` and the candidate is a total, so `lookup()` now returns `None`. Verified by executing both versions: `get_standard_concept('us-gaap_LiabilitiesAndEquity')` returns `'LiabilitiesAndEquity'` on main, `None` on this branch — so "Total Liabilities and Equity" rows silently lose standardization on virtually every balance sheet, contrary to the PR description's "canonical mappings untouched". The regression tests check Assets/AssetsCurrent/Liabilities but not LiabilitiesAndEquity. **Fix shape:** exempt self-matches (tag == candidate) from the floor.

### 2. 🔴 Industry overrides resurrect the suppressed mappings on the mainline path (`reverse_index.py:296`)

The guard evaluates the entry **after** `_apply_industry_override` (lines 290–291), and 29 of the 59 suppressed entries carry industry overrides that are the same total-mapping at high within-industry confidence: `SubordinatedDebt` [Banks]→`Liabilities` 0.961, `SecuredDebt` [Banks/Fin/RlEst]→`Liabilities` 0.85–1.0, `UnearnedPremiums` [Insur]→`Liabilities` 0.993. Since `filing.xbrl()` sets industry from the filer's SIC (`xbrl.py:676` → `cache.py:157` → `core.py:1069`), bank/insurer filings — the exact habitat of #914 — still mis-match. The new tests pass only because they call `get_standard_concept(tag)` with no `industry`. **Fix shape:** apply the floor to override-resolved entries too, and re-run the regression tags with `industry` set.

### 3. 🟡 Survival test asserts the opposite of its name (`tests/issues/regression/test_issue_914_total_false_matches.py:162`)

`test_non_total_low_confidence_mapping_survives` says non-total low-confidence mappings "keep resolving", but `AdministrativeFeesExpense` maps only to `TotalOperatingExpenses` — an all-total mapping the guard suppresses (verified: it now returns `None`). No test exercises a low-confidence non-total tag, so over-suppression regressions pass the suite. Also, every assertion is a pure inequality (`!= total`) that passes for `None` or any wrong concept — the suppressed tags should assert `is None` explicitly.

### 4. 🟡 `_TOTAL_CONCEPT_IDS` has no invariant test (`reverse_index.py:38`)

The set is hand-mirrored from today's 59 rows, and `gaap_mappings.json` is regenerated externally (edgar-storage). A regenerated file with a new low-confidence all-total mapping whose concept id lacks the `Total` substring (e.g. `NetAssets`) slips past `_is_total_concept` with zero signal. Cheap pin: assert every sub-floor all-total entry in the shipped JSON is caught by `_is_total_concept`.

### 5–7. Minor

- **Per-lookup work that's static** (`reverse_index.py:313`): the suppressed set is fully determined at load time; precompute it in `__init__` (the `_embedded_display_names` pattern at lines 141–147) or drop the entries from `_index` at build.
- **Duplicate total heuristic** (`reverse_index.py:52` vs `:407`): `_is_total_concept` (case-sensitive `'Total' in ...` + whitelist) and `_disambiguate_by_context` (`'total' in candidate.lower()`) can drift; consolidate.
- **Dead `isinstance(entry, dict)` checks** (`reverse_index.py:315`): `entry.get(...)` is called unconditionally three lines earlier, so the non-dict branch is unreachable.

ryanduguid:
Addressed all seven findings in `5a79b11a`:

1. Canonical low-confidence self-matches are exempt from suppression. `LiabilitiesAndEquity` now resolves to itself and has a mutation-checked regression test.
2. Suppression is classified from the base mapping before industry confidence can alter it, so Banks, Fin, RlEst, Insur and IFRS industry paths cannot revive these false totals.
3. Suppressed cases now assert `is None`; the non-total survivor asserts the exact `AccountsReceivableGrossNoncurrent -> OtherOperatingNonCurrentAssets` outcome.
4. A shipped-data invariant checks every sub-floor mapping whose candidate display names are totals through the public lookup API, including the canonical self-match exception.
5. Suppressed mapping keys are precomputed once in `ReverseIndex.__init__`.
6. Context disambiguation now uses the shared `_is_total_concept` heuristic.
7. The unreachable lookup-time `isinstance(entry, dict)` checks are gone.

Verification: 56 focused standardisation tests pass; Pyright reports 0 errors; compile and `git diff --check` pass. The full fast suite reports 5,911 passed and 22 unrelated failures; the same 22 node IDs fail on pristine PR-base commit `3e910f8` with matching failure modes on this Windows environment.

PR #1137, opened 2026-08-25
fix(mcp): 13F fund_portfolio lists its holdings again

### LINKED ISSUE #1136: MCP `edgar_ownership(analysis_type="fund_portfolio")` always returns `holdings: []`
## Summary

`edgar_ownership(analysis_type="fund_portfolio")` returns `"holdings": []` for every
fund. `holdings_count` is populated correctly, so the response looks structurally
valid — an agent consuming it concludes the fund reports no positions.

## Reproduction

edgartools 5.52.0 (also reproduces on `main` at aa6dcb04):

```python
import asyncio
from edgar import set_identity
from edgar.ai.mcp.tools.ownership import edgar_ownership

r = asyncio.run(edgar_ownership(identifier="BRK-A", analysis_type="fund_portfolio", limit=3))
print(r.data)
```

Actual, against Berkshire Hathaway's Q2 2026 13F-HR (`0001193125-26-352200`):

```json
{"fund": "BERKSHIRE HATHAWAY INC", "cik": "1067983", "analysis": "fund_portfolio",
 "filing_date": "2026-08-14", "accession_number": "0001193125-26-352200",
 "holdings_count": 29, "holdings": []}
```

Expected: the three largest positions in that filing are Apple (CUSIP 037833100,
227,917,808 shares, $65,950,296,923), American Express (025816109, 151,610,700,
$51,282,319,275) and Coca-Cola (191216100, 400,000,000, $32,508,000,000).

## Cause

`_get_fund_holdings` in `edgar/ai/mcp/tools/ownership.py`:

```python
for h in obj.holdings[:limit]:
    holding = {}
    if hasattr(h, 'cusip'):
        holding["cusip"] = h.cusip
    ...
    if holding:
        holdings.append(holding)
```

`ThirteenF.holdings` is a `pandas.DataFrame`. Iterating a DataFrame yields its
**column names** (`'Issuer'`, `'Class'`, `'Cusip'`, …) as strings, so every
`hasattr` probe is False, `holding` stays empty, and nothing is ever appended.
`holdings_count` is right because it reads `len(obj.holdings)` instead of
iterating.

`_get_portfolio_diff`, in the same file, already walks its DataFrame the correct
way (`for _, row in df.head(limit).iterrows()` with `row.get(...)`), so the fix is
to bring `_get_fund_holdings` into line with its sibling.

One wrinkle worth noting for the fix: the cells come back as numpy scalars, which
`json.dumps` refuses, and as NaN when missing — both need coercing before the
result is serialised over the wire.

## Environment

- edgartools 5.52.0 and `main` @ aa6dcb04
- Python 3.12, macOS

I have a fix and a regression test ready; PR to follow.

hashd1ve:
Fixes #1136.

## The bug

`edgar_ownership(analysis_type="fund_portfolio")` returned `"holdings": []` for
every fund, while `holdings_count` was populated correctly — so the response
looked structurally valid and an agent consuming it concluded the fund reports no
positions.

## Cause

`_get_fund_holdings` iterated `ThirteenF.holdings` and probed each element:

```python
for h in obj.holdings[:limit]:
    if hasattr(h, 'cusip'):
        holding["cusip"] = h.cusip
```

`ThirteenF.holdings` is a `pandas.DataFrame`. Iterating one yields its **column
names** as strings, so every `hasattr` probe was False, `holding` stayed empty and
nothing was appended. `holdings_count` escaped because it reads `len(obj.holdings)`
rather than iterating.

## Fix

`_get_portfolio_diff`, in the same file, already walks its DataFrame with
`for _, row in df.head(limit).iterrows()` and `row.get(...)`. `_get_fund_holdings`
now does the same. Two wrinkles that path already has to handle: the cells come
back as numpy scalars, which `json.dumps` refuses, and as NaN when a column is
empty — `_cell_number` coerces both before the result is serialised over the wire.

## Verification

Against Berkshire Hathaway's Q2 2026 13F-HR (`0001193125-26-352200`), fetched
2026-08-25:

```json
{"holdings_count": 29,
 "holdings": [
   {"company": "APPLE INC", "cusip": "037833100", "shares": 227917808, "value": 65950296923},
   {"company": "AMERICAN EXPRESS CO", "cusip": "025816109", "shares": 151610700, "value": 51282319275},
   {"company": "COCA COLA CO", "cusip": "191216100", "shares": 400000000, "value": 32508000000}],
 "total_value_shown": 149740616198}
```

Those three positions are the fixture behind
`tests/issues/regression/test_issue_1136.py`, so the test is a unit test and
records no cassette. It asserts the values, the `limit` behaviour and that the
payload survives `json.dumps`; on `main` it fails with `holdings == []`.

`hatch run check-regression-provenance` passes (303 files). `ruff check` on the
touched file goes from 5 findings to 4 — the previously unused `Optional`/`Any`
imports are now used.

I found this while evaluating an agent against the MCP server: the model was asked
how many securities Berkshire reports and had no way to get them from this tool.


dgunning:
## Code review — core fix verified correct, 2 minor findings

The fix holds up under verification: column names (`Issuer`, `Cusip`, `SharesPrnAmount`, `Value`) were checked against both real 13F parsers (`thirteenf/models.py` and the XML/TXT infotable parsers), the `is not None` guard correctly replaces the always-true `hasattr` probe for the `None`-returning `holdings` cached_property, `head(limit)` and `holdings_count` preserve the old semantics, and the regression tests pass (run in a clean worktree). A suspected `pd.NA` crash was probed against the PR head and refuted — keys are cleanly omitted, no error.

### 1. 🟡 NaN Issuer/Cusip cells serialize as the literal string `"nan"` (`edgar/ai/mcp/tools/ownership.py:252`)

`bool(float('nan'))` is `True`, so `if issuer:` / `if cusip:` admit NaN cells and `str(...)` emits `{"company": "nan", "cusip": "nan"}` — confirmed by probe. The XML parser defaults these to `""` so the trigger needs a degraded TXT-format parse, but the sibling `_get_portfolio_diff` already guards its Ticker with `str(...) != "nan"`. A `pd.notna(issuer)` check closes it.

### 2. 🟡 Duplicate NaN/numpy coercion helpers that disagree (`ownership.py:95`)

The new `_cell_number` duplicates the existing `_safe_number` in `_get_portfolio_diff` (same module, line ~343 — which is also re-defined inside the per-row loop on every iteration). They disagree on int coercion (`_safe_number` casts to int, `_cell_number` returns floats as-is), so the two analysis types can render the same Value column differently. Hoist one module-level helper and use it in both walks.

Fine to merge as-is with these folded in or handled as a small follow-up.

PR #1139, opened 2026-08-25
chore(markdown): drop three dead members from edgar/_markdown.py (07lk.12.1)

dgunning:
Deletion pass ahead of the `_markdown` → `richtools` consolidation in `07lk.12.1`. Nothing removed here was reachable. The module goes from **139 lines to 104**, net −74 across the diff.

## What was dead, and since when

| member | last caller | status |
|---|---|---|
| `fix_markdown()` | **none, ever** | dead the whole time — not production, not internal, only its own test |
| `MarkdownContent.from_html()` | `PressRelease.to_markdown()` | died **this release** in #1132 |
| `html_to_markdown()` | `MarkdownContent.from_html()` | died with it |

`from_html` was `html_to_markdown`'s only caller, and its own last caller went when press releases moved to `Attachment.markdown()` and off the legacy parser. Both then survived only in two tests exercising them for their own sake.

`edgar/_markdown.py` is private and appears in no documentation, so nothing reachable through the public API changes — the same situation as the `edgar.datatools.table_tag_to_dataframe()` removal already in this release's notes.

## What remains

Exactly what has callers:

- `convert_table` → `edgar/search/textsearch.py:220`
- `MarkdownContent` → `edgar/company_reports/press_release.py:88`
- `markdown_to_rich` → internal, via `MarkdownContent.__rich__`
- `text_to_markdown` → `edgar/_filings.py:1813`

## Tests

The two `from_html` tests are **rewritten rather than deleted** — they covered `MarkdownContent.__rich__` against real 6-K HTML, which is worth keeping. They now build the markdown through `edgar.documents` and wrap it, which is how the remaining callers construct it. `test_fix_markdown` is deleted outright, there being nothing left to test.

Also drops an unused `Document` import and sorts the import block in `tests/test_markdown.py`. Both were pre-existing — ruff reports identically on main — but a dead import belongs in a dead-code pass, in a file this change is already editing.

## Verification

**6343 pass in both lanes** — `test-fast` and `EDGARTOOLS_STRICT_ERRORS=1 pytest -n auto -m 'fast'` — one fewer than main, which is exactly `test_fix_markdown`.

The three surviving exports exercised through their real callers rather than only their own tests: `tests/test_textsearch.py` 15 pass (`convert_table`), `tests/test_eightK.py` 19 pass (`MarkdownContent` via `PressRelease`). ruff clean on both changed files.


dgunning:
Added a second commit (2d8ebe0d) that reconciles a contradiction #1135 introduced into the same release notes.

The `### Removed` entry for the 3dp Group B work said:

> A lookup for an item a filing does not have now returns `None` from every report type

True when written, and no longer true after #1135 — those lookups warn now, and raise `SectionNotFoundError` under `EDGARTOOLS_STRICT_ERRORS`. Both bullets ship in the same release, so a reader would have hit two different answers to the same question. The bullet now states current behaviour, keeps the historical point that `TenK` used to raise `TypeError` there, and points at the Fixed entry covering it.

I scanned the rest of `[Unreleased]` for the same staleness rather than fixing only the one I noticed. The other item-lookup bullet ("a 10-K item lookup spelled in capitals came back empty") describes the old behaviour in the past tense and is still accurate; this was the only stale one. Bullet count unchanged at 481 — a modification, not an addition.

It rides on this branch rather than its own deliberately: both edits are inside the same `### Removed` block, and `CHANGELOG.md` carries the `union` merge driver. Two branches editing overlapping hunks there merge by keeping **both** sides — which is precisely how a contradictory changelog gets written.

PR #1142, opened 2026-08-25
fix(mcp): one cell reader for both 13F walks

hashd1ve:
Follow-up to the review on #1137, which left two findings open. Both are in `edgar/ai/mcp/tools/ownership.py`, and the second is why the first had to be fixed twice.

## A position the parser could not name was reported as `"nan"`

A missing text cell arrives as NaN, and NaN is truthy. `if issuer:` let it through and `str(nan)` is the literal `"nan"`, so `fund_portfolio` answered:

```json
{"company": "nan", "cusip": "nan", "shares": 400000000, "value": null}
```

The 13F XML parser defaults those columns to `""`, so the trigger is a degraded TXT-format parse. `_get_portfolio_diff` was already guarding its `Ticker` with `str(...) != "nan"` for exactly this — and guarding nothing else: its `issuer`, `cusip` and `status` passed the bare NaN straight through, which `json.dumps` writes as the literal `NaN`. That is not JSON a strict parser reads back.

## The same value, rendered two ways

`_cell_number` and `_safe_number` did the same coercion in the same module — the latter redefined inside the per-row loop, re-importing pandas on every iteration — and disagreed about int coercion. One missing cell makes its whole column float64, so Berkshire's Apple stake came back as `65950296923.0` from `fund_portfolio` beside `65950296923` from `portfolio_diff`: same number, same filing, same quarter.

## What changed

Both walks now read cells through two module-level helpers:

- `_cell_text` — omits a missing cell instead of naming it `"nan"`, and strips.
- `_cell_number(value, as_int=True)` — keeps whole-number columns whole, so both walks agree; `share_change_pct` passes `as_int=False` and stays fractional.
- `_cell_missing` behind them uses `pd.isna`, which also recognises `pd.NA` — `value != value` raises on that one.

One behaviour change worth naming: a cell that is not a number at all is now handed back rather than raising `ValueError` out of the tool.

## Verification

Ground truth is again Berkshire Hathaway's Q2 2026 13F-HR (`0001193125-26-352200`), with a third row standing in for a parse that recovered the numbers but not the issuer — the shape that puts NaN in a text column and floats in a numeric one. The frames are built from recorded values, so the tests need no network.

- 6 new tests in `tests/issues/regression/test_mcp_ownership_cell_coercion.py`; **5 fail on main** (`'nan' != None`, `isinstance(65950296923.0, int)`, `nan == ''`, `Out of range float values are not JSON compliant: nan`, `float is not int`). The sixth guards the fractional column against the unification.
- `test_issue_1136.py` still passes unchanged.
- `ruff check` reports the same four pre-existing findings on the file as before, and no new one.
- `tests/test_mcp_intent_tools.py` fails 14 tests here — identically on main, they are the network ones.


dgunning:
## Code review — 6 findings (verified against the branch)

The core coercion fix works, but this needs another pass before merge: it re-emits the bug from its own before-picture in one shape, and the same NaN defect it fixes is still live in sibling tools.

### 1. 🔴 `fund_portfolio` still emits the explicit `"value": null` from the PR's own bug example (`edgar/ai/mcp/tools/ownership.py:294`)

`shares`/`value` are written as a pair when *either* is present: `if shares is not None or value is not None:` stores `None` for the missing half, while `company`/`cusip` (and every `_get_portfolio_diff` numeric) omit the key. The unparsed row (shares=400000000, Value=NaN) yields `{"shares": 400000000, "value": null}` — the same `"value": null` shown in the PR description's before-picture. A caller using `"value" in holding` to detect presence gets `True` with `None`. **Fix:** two independent `if x is not None` guards; add a test for the one-of-the-pair-missing shape (none covers it).

### 2. 🔴 The fix stops one file too early — the identical truthy-NaN bug is live in sibling tools, and the helpers belong in `tools/base.py` (`edgar/ai/mcp/tools/screen.py:136`)

`screen.py:136` `"name": row.get('name') or row.get('company', '')` lets a bare NaN through (NaN is truthy), and `ToolResponse.to_json` (plain `json.dumps`, no `allow_nan=False`) writes literal `NaN` — the exact invalid-JSON defect this PR fixes in `portfolio_diff`; screen.py even hand-rolls `str(x) != 'nan'` guards for ticker/exchange three lines below and still missed `name`. `search.py:129` `if row.ticker:` has the same trap, and `fund.py:27-51` / `proxy.py:42-65` carry further near-duplicate `_df_to_records` NaN coercers. The new helpers are the third private copy of cell coercion in `edgar/ai/mcp/tools/` — hoisting `_cell_missing`/`_cell_text`/`_cell_number` into `tools/base.py` fixes the live siblings and stops the per-tool drift.

### 3. 🟡 `test_fractional_columns_stay_fractional` cannot fail for the regression it guards (`tests/issues/regression/test_mcp_ownership_cell_coercion.py:167`)

`int == float` in Python: if `as_int=True` were wrongly applied to `ShareChangePct`, `_cell_number(-24.03)` → `-24`, `round(-24, 1)` → `-24`, and `assert -24 == -24.0` **passes** — i.e. the test passes under exactly the unification it claims to guard against. **Fix:** assert `isinstance(share_change_pct, float)`, or use a fixture value like `-24.5` whose truncation changes the number.

### 4. 🟡 The shares half of the int64 test exercises nothing (`test_mcp_ownership_cell_coercion.py:229`)

The fixture comment says the missing cells make both `SharesPrnAmount` and `Value` float64, but only `Value` has a gap — row 3 has `SharesPrnAmount=400000000`, so that column stays int64 and `isinstance(apple['shares'], int)` passes even with the fix reverted. Give row 3 a missing `SharesPrnAmount` (it already claims "recovered the numbers but not the issuer") or drop the inert assertion.

### 5. 🟡 `as_int: bool = True` makes silent truncation the default (`ownership.py:123`)

No current column loses data (Shares/Value are int-at-source; floats arise only from NaN promotion — verified), but the next genuinely fractional column wired through the default is silently floored via `int(value)`, and finding 3's inert guard means no test would catch it. Invert the default: `as_int=False`, with the six count/dollar call sites passing `True` explicitly.

### 6. ⚪ Changelog entry: 153 words vs the 80-word cap, wrong ending (`CHANGELOG.md:49`)

CONTRIBUTING.md:155 caps the whole entry at eighty words and :165 requires ending with `(GH #NNN)` and nothing else; this entry closes with "(Follow-up to the review on #1137)". Trim the mechanism detail (it lives in the PR body and test docstrings) and end with `(GH #1137)`.

Verified along the way (no action needed): `Shares`/`Value` are int-at-source across both 13F parsers, so no real fractional data is truncated today; the non-numeric handed-back behavior change is acknowledged in the PR body.

hashd1ve:
All six confirmed and fixed. Thank you — findings 1 and 2 in particular were things I saw and talked myself out of.

**1 — the explicit `"value": null`.** Right, and it was in my own before-picture. `shares` and `value` are written independently now, the way `company` and `cusip` already were. I had read the paired write as deliberate and left it; it isn't defensible next to the very defect the PR fixes.

**2 — one file too early.** The helpers are in `edgar/ai/mcp/tools/base.py`, and the live siblings go through them:

- `screen.py:136` — confirmed: `bool(nan)` is `True`, so a NaN name passed through and `to_json` (plain `json.dumps`, no `allow_nan=False`) wrote it as literal `NaN`.
- `search.py:129` — same trap on `if row.ticker:`.
- `fund.py` and `proxy.py` — both `_df_to_records` copies now call `_cell_missing`. `_cell_number` learned `Decimal` while I was there, since that is what those two convert by hand and what NPORT columns arrive as.

`ToolResponse.to_json` would catch the next one of these at the boundary if it passed `allow_nan=False` — an exception instead of invalid JSON on the wire. I did not do it here: it changes failure behaviour for every tool at once, so it seemed yours to decide rather than something to fold into this PR.

**3 — the inert fractional test.** Verified: `int(-24.03)` is -24, `round(-24, 1)` is -24, and `-24 == -24.0` passes. The fixture is `-24.5` now, whose truncation changes the number, plus an `isinstance(..., float)` assertion.

**4 — the inert int64 assertion.** Also right: only `Value` had a gap, so `SharesPrnAmount` stayed int64. A fourth row loses its share count instead of its value, so both columns arrive float64 — the test asserts the dtypes directly rather than trusting the comment — and that row covers the other half of finding 1.

**5 — `as_int` default.** Inverted to `False`; the seven whole-number call sites pass `True` explicitly.

**6 — changelog.** 80 words, ends `(GH #1137)`. I have trimmed the entries on #1144 and #1147 to the same rule.

Evidence that each new test bites: reverting only `screen.py` and `search.py` fails the two sibling tests on `assert nan == ''` and `'ticker': nan`; reverting only `ownership.py` fails the pair test on `{'shares': 400000000, 'value': None}`. 18 tests in the file, 100 across the four MCP suites, and ruff reports the same 31 findings across the six touched modules as `main` — the same set, not just the same count.

Issue #888, opened 2026-07-07
Fund.get_filings() with series_only=True fails to filter filings by the fund's series_id


HristoRaykov:
## Data Quality Issue Details

**Issue Type:**
- [x] Incorrect financial values (wrong numbers)
- [ ] Missing financial data (expected data not present)
- [ ] Calculation errors (formulas producing wrong results)
- [ ] Data inconsistency (different values for same metric)
- [ ] Historical data problems (changes over time)

## Environment
**EdgarTools Version:** 5.40.1
**Python Version:** 3.14.3 
**Operating System:** Windows 11

## Financial Data Details
**Company/Ticker:** VCLT, PFF, ...
**Form Type:** NPORT-P

**Code to reproduce:**
```python

from edgar import *

set_identity('')

fund = Fund("VCLT") # PFF ...

filings = fund.get_filings(form=['N-PORT', 'NPORT-P'], series_only=True)
filing = filings[0]
report = filing.obj()
print(f'Fund Series ID: {fund.series.series_id}')
print(f'Report Series ID: {report.series_id}')
print(f'Filing URL: {filing.url}')
#  latest company NPORT filing matches report filing
print(f'Company Filings: https://www.sec.gov/edgar/browse/?CIK={fund.company.cik}')

portfolio = report.investment_data()

```

## Expected Behavior
The get_filings(..., series_only=True) method should filter the returned Filings object so that all underlying filings have a series_id that perfectly matches fund.series.series_id.

## Actual Behavior
The method returns latest NPORT-P filing belonging to the broader company/trust, and the report.series_id from the fetched filing does not match the fund.series.series_id.

dgunning:
Thanks for the clear report and reproduction — fixed on `main` (0d188b21).

You diagnosed the symptom exactly: `series_only=True` was returning the umbrella trust's filings, so `report.series_id` didn't match `fund.series.series_id`. Root cause was that the series-aware path searched EDGAR full-text search (EFTS) for the series ID string — which returns nothing, because SEC full-text search doesn't index NPORT series IDs. That empty result caused a silent fall-through to the unfiltered trust delegation, so you got a sibling series' data (VCLT / S000026864 was returning a filing for S000026865).

**Fix:** `series_only=True` now resolves the series through SEC browse-edgar using the series ID as the CIK parameter, which returns exactly that series' filings. And when a series has no matching filings, it now returns an empty `Filings` rather than silently falling back to the trust.

Verified on your examples:
[...]

PR #1144, opened 2026-08-25
fix(funds): a series answers with its own filings, not its trust's

### LINKED ISSUE #1143: `FundSeries.get_filings()` returns the whole trust's filings, so a series' newest NPORT-P is a sibling's
`FundSeries.get_filings()` delegates straight to the fund company:

```python
# edgar/funds/core.py
def get_filings(self, **kwargs) -> 'Filings':
    return self.fund_company.get_filings(**kwargs)
```

A trust files one NPORT-P per series per quarter under a single CIK, so that returns every sibling series' filings mixed together, and the newest belongs to whichever series filed last — not to the series that was asked. `docs/examples/user_journeys.md:66` recommends exactly this call:

```python
nport_filings = series.get_filings(form='NPORT-P')
```

## Reproduction

edgartools 5.52.0. Two series of Vanguard's trust, CIK 36405:

```python
from edgar.funds import find_fund

for ticker in ("VEXMX", "VFINX"):
    series = find_fund(ticker).series
    filings = series.get_filings(form="NPORT-P")
    print(ticker, series.series_id, len(filings), filings[0].accession_number)
```

```
VEXMX S000002841 325 0000036405-26-000325
VFINX S000002839 325 0000036405-26-000325
```

Same count, same filing. `0000036405-26-000325` has `general_info.series_id == "S000002839"`, so VEXMX — Vanguard Extended Market Index, which **excludes the S&P 500 by construction** — is handed the 500 Index fund's portfolio, with NVDA at 7.58%.

Expected: each series' own filings (28 and 27 respectively, topped by `...316` and `...325`).

What makes this one unpleasant is that nothing about the answer looks wrong. The accession is real, the holdings are real, the numbers are real — they are just the wrong fund's. Any provenance check downstream passes.

## Cause, and its relation to #888

This is #888's trap on the neighbouring object. `Fund.get_filings(series_only=True)` was fixed there by resolving the series through browse-edgar with the series ID in the CIK slot, and by settling that an unresolvable series must return empty rather than the trust, because "returning trust-wide filings here would give the caller a sibling series' data". `FundSeries.get_filings` never got that treatment, and it is the path a `FundSeries` object naturally offers.

The two are also connected in the other direction: `Fund.get_filings()`'s default path reaches the trust *through* `FundSeries.get_filings` (`self._entity` is a `FundClass` for a ticker, which has no `get_filings`, so it falls to `self._series.get_filings`).

## A question that is yours, not mine

`Fund.get_filings()` documents its default as the umbrella trust's filings, with `series_only=True` as the narrowing ask, and `test_default_get_filings_unchanged` pins that. So for a ticker that resolves to exactly one series, `Fund("VEXMX").get_filings(form="NPORT-P")[0]` is still the 500 Index report by design.

Is that still the intent? The argument for keeping it: the series path costs a browse-edgar lookup per requested form, and an unfiltered ask on a large trust otherwise pages through its whole history (the 503 risk noted in `_get_series_filings`). The argument against: a caller who resolved a fund by *ticker* has named one series, and the trust-wide answer is wrong for the same reason it was wrong in #888.

PR to follow fixes `FundSeries` only and leaves that default alone, so the decision stays open.

hashd1ve:
Fixes #1143.

`FundSeries.get_filings()` was a delegation to `self.fund_company.get_filings(**kwargs)`. A trust files one NPORT-P per series per quarter under a single CIK, so that returns every sibling series' filings mixed together and the newest belongs to whichever series filed last.

Vanguard's trust (CIK 36405), on 5.52.0:

```
VEXMX  S000002841  325 filings  top: 0000036405-26-000325
VFINX  S000002839  325 filings  top: 0000036405-26-000325
```

That accession's `general_info.series_id` is `S000002839`, so `series.get_filings(form="NPORT-P")[0]` — the call `docs/examples/user_journeys.md:66` recommends — gave Vanguard Extended Market Index, which excludes the S&P 500 by construction, the 500 Index fund's portfolio. Real accession, real holdings, real numbers, wrong fund.

## The fix

This is #888's trap on the neighbouring object, so it takes #888's path. The body of `Fund._get_series_filings` moves to a module-level `_series_filings()`; `Fund._get_series_filings` stays as a thin delegate, and `FundSeries.get_filings` now calls it too. An unresolvable series yields no filings rather than the trust's, for the reason #888 settled: an empty answer is correct for a series with nothing on file, a sibling's data never is. An ETF's synthetic `ETF_<cik>` series stands for the whole registrant, has no sibling to be confused with, and still asks its company.

After:

```
VEXMX  S000002841   28 filings  top: 0000036405-26-000316  (general_info.series_id S000002841)
VFINX  S000002839   27 filings  top: 0000036405-26-000325  (general_info.series_id S000002839)
```

## What is deliberately *not* changed

`Fund.get_filings()`'s documented trust-wide default. It reached the trust *through* `FundSeries.get_filings`, so it would have changed silently; the default path now routes a series entity to its fund company explicitly and answers exactly as before. `series_only=True` remains the narrowing ask and remains the only path that pays for a browse-edgar lookup per requested form — which also keeps the unfiltered-large-trust 503 risk noted in `_get_series_filings` off the default path.

Whether that default should change at all is the open question in #1143, and it is yours; this PR leaves it open.

## Verification

- 6 new tests in `tests/issues/regression/test_fund_series_get_filings_own_series.py`: three offline routing tests, two network tests asserting the returned report's `general_info.series_id` is the series that was asked, and one pinning the trust-wide default. **Four fail on main**, the network one on `assert 'S000002839' == 'S000002841'`.
- The #888 suite passes unchanged, `test_default_get_filings_unchanged` included.
- 276 passed / 1 skipped across the eleven `tests/*fund*.py` suites.
- `ruff check` reports the same three pre-existing findings on `edgar/funds/core.py` and no new one.


dgunning:
Apologies — this PR was missed, and a duplicate fix (#1162) was written and merged over it. That was my error: I triaged #1143 without checking for an existing PR. #1162 has been reverted in #1165 so this lands as the fix for #1143.

Your diagnosis and design were right, and independently matched: the module-level series resolver, browse-edgar with the series ID in the CIK slot, the synthetic `ETF_*` guard, and refusing to fall back to the trust for the reason #888 settled. Nothing here needs changing on that account.

Three things the duplicate carried that this PR doesn't cover, which I'll send as a follow-up on top of yours rather than asking you to expand scope:

- `get_latest_report()` and `get_portfolio()` chain through `get_filings()` without `series_only`, so `Fund("VEXMX").get_latest_report()` still returns a sibling's report — the same harm through the documented holdings entry point.
- `year`/`quarter` as a list is untranslatable in `_series_filter_kwargs`, silently dropped, so the caller gets the series' entire unfiltered history.
- `Fund.get_filings()`'s trust-wide default reaches the series entity via `_entity`/`_series`, so it needs routing to the company explicitly or the default quietly becomes series-scoped.

Thanks for the fix, and sorry for the detour.

Issue #912, opened 2026-07-25
497K parsing issues


dima:
## Issue Type

- [х] Data Quality Issue (incorrect financial values, missing data)
- [ ] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.43.0
**Python Version:** 3.14.3
**Operating System:** MacOS 15.7.8

## Bug Description
**What you expected to happen:**

I expected the following:
```
    from edgar import find

    filing = find("0001580642-24-004234")
    p = filing.obj()
    print(p)
```

to correctly parse the given 497K filing as per documentation here: https://edgartools.readthedocs.io/en/latest/guides/prospectus497k-data-object-guide/?h=497

If you open the filing there's info on tickers, expenses, classes/series and so on, which is not reflected in parsed output.

I am sure this is not the case with *all* 497K reports but in this case there's a problem.

https://www.sec.gov/Archives/edgar/data/1314414/000158064224004234/0001580642-24-004234-index.html

**What actually happened:**
>>> filing = find("0001580642-24-004234")
>>> p = filing.obj()
>>> print(p)
Prospectus497K(Ocean Park High Income ETF, 2 classes, no tickers)

>>> p.tickers
[]                                                                               
>>> p.expense_example                                                                               
class_name ticker  1yr   3yr   5yr  10yr
0  Ocean Park High Income ETF   None    
3  None  None  None                      1                        
(3)   None    3  None  None  None

---
*This issue will be handled using EdgarTools' systematic issue resolution workflow. A reproduction test will be created to verify the fix.*

dgunning:
Status update — **partially fixed in 5.43.1**, keeping this open for the rest.

## Fixed and released (5.43.1)

Two defects in the fee-table numbers, both reproduced on your Ocean Park filing:

1. **Fee waiver held the net expense ratio.** For tables using the standard "After Fee Waiver and Reimbursement" wording, `fee_waiver` was populated with the net expense ratio and `net_expenses` was left `None`. Net expenses are now matched first and the waiver is normalised to a signed reduction, so `total_annual_expenses + fee_waiver == net_expenses` holds.
[...]

PR #1160, opened 2026-08-30
fix: distinguish 497K fee tables from footnotes

sambai-dev:
## Problem

A 497K fee-waiver footnote can be rendered as its own one-row table. The Ocean Park filing in #912 includes the phrase "management fees" in that prose, so `_classify_table` treated the footnote as a second operating-expenses table.

That false positive changed `extract_fee_tables` to the repeated-section path. The single-column expense parser then read `3 Years` as `$3`, returned no three-year value, and emitted the footnote as a phantom share class. The reduced filing excerpt returned two records with `expense_1yr == 3` instead of one record with expenses of $109 and $381.

## Approach

- Require fee-label/value structure before classifying a table as operating expenses. Missing values with conventional footnote markers and integer values remain valid, while percent-bearing narrative must match the complete SEC value shape.
- Stop prose-only management-fee tables before they fall through into expense-example, shareholder-fee, performance, or bar-chart classifiers. Mixed tables retain those classifications only when they contain the corresponding structural signal.
- Normalize a shared leading spacer offset across the whole operating-expense or shareholder-fee table so multi-class headers and values remain aligned.
- Preserve malformed nested tables that flatten into one cell only when the label is immediately followed by multiple percentage values, including the existing trailing-empty-cell shape.
- Add a reduced regression fixture preserving the three relevant Ocean Park tables, plus end-to-end cases for marked missing values, percent-bearing prose, single- and multi-class spacer layouts, flattened rows, and indented shareholder fees.
- Update the existing 497K corpus baseline to remove the three phantom records created from one-row prose footnotes.

I rejected filtering malformed classes after extraction because the false table count has already selected the wrong expense-example parser by then. I also rejected a literal prefix check because valid labels can be prefixed or footnote-marked, while prose can itself begin with "management fees."

A corpus-wide classifier comparison changed exactly three tables, all one-row prose footnotes. Existing management-fee label rows remained classified as operating expenses.

## Review follow-up

The maintainer-reported fall-through, marked-value, spacer-column, and flattened-row cases are covered directly. Follow-up review also added end-to-end protection for percent-leading narrative, multi-class value alignment, and shareholder-fee extraction after indentation normalization.

## Compatibility

This does not change the public API. It narrows an internal classifier false positive and preserves existing SEC value forms, including leading-decimal percentages, split-text percentages, parenthesized values, numeric or symbol markers, and single-letter marker lists.

This intentionally leaves the cover-page ticker fallback from #912 out of scope because it is a separate extraction path with different fallback rules.

## Testing

- Every reported and review-discovered behavior was reproduced as an isolated failing test before its source fix.
- `uvx hatch run pytest tests/issues/regression/test_issue_912_497k_fee_waiver.py tests/test_prospectus497k.py -q --tb=short`: 98 passed.
- Focused 497K characterization command: 50 passed with the same three environment-dependent metadata-only baseline cases excluded; the untouched branch fails those same three cases.
- Fast xdist and coverage gate: 6,433 passed; its 23 failures exactly match the untouched branch.
- Full regression gate: 2,950 passed; its 10 failures exactly match the untouched branch.
- Cassette safety, regression no-skip, regression provenance, Ruff, Pyright, added-line security, and whitespace checks introduced no new findings.
- The exact final diff and its current-base integration passed a fresh fail-closed review with no logic, security, scope, or test-gap findings.

Related to #912

### DIFF
--- a/edgar/funds/_497k_tables.py
+++ b/edgar/funds/_497k_tables.py
@@ -177,9 +177,9 @@ def _label_matches(text: str, *patterns: str) -> bool:
 # Table classification
 # ---------------------------------------------------------------------------
 
-_FEE_TABLE_LABELS = (
-    'management fee',
-    'management fees',
+_MANAGEMENT_FEE_LABEL_RE = re.compile(r'\bmanagement fees?\b')
+_FLATTENED_FEE_VALUES_RE = re.compile(
+    r'^(?:\([^)]*\)\s*)?(?:[-+]?\(?\d*\.?\d+\)?\s*%\s*){2}'
 )
 
 _EXPENSE_EXAMPLE_LABELS = (
@@ -211,6 +211,29 @@ _BAR_CHART_LABELS = (
 )
 
 
+def _is_management_fee_row(row: List[str]) -> bool:
+    """Return whether a row has management-fee label/value structure."""
+    if not row:
+        return False
+
+    label = _normalize(row[0])
+    match = _MANAGEMENT_FEE_LABEL_RE.search(label)
+    if match is None:
+        return False
+    if len(row) > 1:
+        return any(
+            _normalize(cell) in {'none', 'n/a', 'na', 'not applicable', '-', '--', '–', '—'}
+            or '%' in cell
+            or (_parse_percentage(cell) is not None and '.' in cell)
+            for cell in row[1:]
+        )
+
+    # Some malformed nested tables flatten all fee rows into one cell. Keep
+    # those only when the label is immediately followed by two percentage values.
+    trailing = label[match.end():].lstrip()
+    return match.start() == 0 and _FLATTENED_FEE_VALUES_RE.match(trailing) is not None
+
+
 def _classify_table(rows: List[List[str]]) -> Optional[str]:
     """Classify a table by its content. Returns a type string or None."""
     all_text = ' '.join(' '.join(row) for row in rows)
@@ -225,8 +248,9 @@ def _classify_table(rows: List[List[str]]) -> Optional[str]:
             ('quarter' in norm or 'return' in norm)):
         return 'quarter'
 
-    # Check for operating expenses table (has "management fee")
-    if any(p in norm for p in _FEE_TABLE_LABELS):
+    # Match the same label/value structure consumed by the fee extractor so
+    # prose-only footnote tables do not change the extraction strategy.
+    if any(_is_management_fee_row(row) for row in rows):
         return 'operating_expenses'
 
     # Tables with year-period columns (1 year, 3 years, etc.)
--- a/edgar/funds/_497k_tables.py
+++ b/edgar/funds/_497k_tables.py
@@ -178,9 +178,27 @@ def _label_matches(text: str, *patterns: str) -> bool:
 # ---------------------------------------------------------------------------
 
 _MANAGEMENT_FEE_LABEL_RE = re.compile(r'\bmanagement fees?\b')
+_ONE_YEAR_CELL_RE = re.compile(r'^(?:past\s+)?1\s*years?$')
 _FLATTENED_FEE_VALUES_RE = re.compile(
     r'^(?:\([^)]*\)\s*)?(?:[-+]?\(?\d*\.?\d+\)?\s*%\s*){2}'
 )
+_FEE_VALUE_FOOTNOTE_RE = re.compile(r'(?:\s*\(\s*[\d*†‡§]+\s*\)|\s*[*†‡§]+)\s*$')
+_PERCENT_VALUE_SHAPE_RE = re.compile(
+    r'''
+    ^\s*\(?\s*[-+]?\s*
+    (?:\d+(?:\.\s*\d*)?|\.\s*\d+)
+    \s*\)?\s*%\s*\)?\s*
+    (?:
+        [\d*†‡§\s]+
+        | \(\s*[\d*†‡§\s]+\s*\)
+        | [A-Za-z](?:\s*,\s*[A-Za-z])*
+        | \(\s*[A-Za-z](?:\s*,\s*[A-Za-z])*\s*\)
+    )?
+    \s*$
+    ''',
+    re.VERBOSE,
+)
+_MISSING_FEE_VALUES = {'none', 'n/a', 'na', 'not applicable', '-', '--', '–', '—'}
 
 _EXPENSE_EXAMPLE_LABELS = (
     '1 year',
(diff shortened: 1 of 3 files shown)


edgar/funds/_497k_tables.py
```diff
@@ -225,8 +248,9 @@ def _classify_table(rows: List[List[str]]) -> Optional[str]:
             ('quarter' in norm or 'return' in norm)):
         return 'quarter'
 
-    # Check for operating expenses table (has "management fee")
-    if any(p in norm for p in _FEE_TABLE_LABELS):
+    # Match the same label/value structure consumed by the fee extractor so
+    # prose-only footnote tables do not change the extraction strategy.
+    if any(_is_management_fee_row(row) for row in rows):
```
dgunning, line 322: **Prose footnotes now fall through into the *other* classifier branches.**

Removing a footnote from the `operating_expenses` bucket doesn't remove it from classification — it keeps walking down this chain and matches `expense_example`, `shareholder_fees` or `performance`, all of which use "first table wins". Reproduced against `main` on identical HTML:

- Footnote whose prose mentions "1 Year" and a `$` amount, placed between the fee table and the real Example table → classified `expense_example`, becomes `expense_ex_tables[0]` (the only one the single-fee-table path reads). `main`: `{... 'expense_1yr': 109, 'expense_3yr': 381}`; PR: those keys are gone. Same class of loss as #912, in the opposite direction.
- Footnote mentioning "sales charge" placed before the shareholder-fee table → classified `shareholder_fees`, becomes `sh_fee_tables[0]`, and `max_sales_load` / `redemption_fee` disappear.
- Already in the corpus: `tests/fixtures/prospectus_497k/2012q1-1493113.html` table 6 (the fee-waiver footnote) is now classified `shareholder_fees`. Harmless in that one file only because it yields no shareholder-fee data either way.
- A two-row footnote mentioning "1 Year" and `%` is now classified `performance` and injects a junk row into `extract_performance_table`: `[{'label': '(2)', 'section': '', 'return_1yr': None}]` (`main` returned `[]`).

Suggested direction: give prose-only tables an explicit terminal classification (return `None` / `'footnote'` when the only signal is management-fee prose) rather than letting them continue down the branch chain.

edgar/funds/_497k_tables.py
```diff
@@ -211,6 +211,29 @@ def _label_matches(text: str, *patterns: str) -> bool:
...
+
+    label = _normalize(row[0])
+    match = _MANAGEMENT_FEE_LABEL_RE.search(label)
+    if match is None:
+        return False
+    if len(row) > 1:
+        return any(
+            _normalize(cell) in {'none', 'n/a', 'na', 'not applicable', '-', '--', '–', '—'}
+            or '%' in cell
+            or (_parse_percentage(cell) is not None and '.' in cell)
```
dgunning: **One unrecognised value cell discards the entire filing's fee extraction.**

A value is accepted only if it is `%`-bearing, a whitelisted none-token, or parses *and* contains `.`. `"None (1)"` — a footnote-marked none, very common typesetting — matches none of these (`_parse_percentage`'s none-whitelist requires the string be exactly `none`). Since this row check is now the sole gate for `operating_expenses`, the table is unclassified, `op_expense_tables` is empty, and `extract_fee_tables` returns `[]`.

Reproduced with a table whose management-fee value is `None (1)` and whose other rows are all valid:
- `main` → `{'twelve_b1_fee': 0.25, 'total_annual_expenses': 0.60, 'max_sales_load': 5.75, 'expense_1yr': 109, 'expense_3yr': 381}`
- PR → `[]`

Same collapse for `— (1)`, `N/A*`, and a bare integer with no `%` and no decimal point (`1`). The predicate should be a weaker "this row is not prose" test, or this row check shouldn't be the only thing that can classify the table.

dgunning (commented):
Automated review of the classifier change. Three findings, all reproduced by running this branch side by side against `main` on identical HTML.

The narrowing correctly stops the #912 footnote from landing in the `operating_expenses` bucket, but a footnote that is no longer `operating_expenses` is not thereby unclassified — it falls through to the later branches, where "first table wins" makes it displace real data. Two of the three findings are that fall-through and the over-tight row predicate that now gates the whole filing.

Non-issues checked: no consumers of the removed `_FEE_TABLE_LABELS` outside this module; `tests/test_497k_tables_characterization.py`, `tests/issues/regression/test_issue_912_497k_fee_waiver.py` and `tests/test_prospectus497k.py` all pass on this branch (104 passed, 28 deselected); the baseline JSON edits are correct — the only non-phantom change (`2016q4-3794`) drops three keys whose values were already `null`.

edgar/funds/_497k_tables.py
```diff
@@ -211,6 +211,29 @@ def _label_matches(text: str, *patterns: str) -> bool:
...
+def _is_management_fee_row(row: List[str]) -> bool:
+    """Return whether a row has management-fee label/value structure."""
+    if not row:
+        return False
+
+    label = _normalize(row[0])
+    match = _MANAGEMENT_FEE_LABEL_RE.search(label)
+    if match is None:
+        return False
+    if len(row) > 1:
```
dgunning: **Requiring the label in `row[0]` drops tables with a leading indent/spacer cell, and the flattened-cell fallback below is unreachable.**

`<tr><td></td><td>Management Fees</td><td>0.65%</td></tr>` — an indentation column, common in SEC HTML — no longer classifies, so `extract_fee_tables` returns `[]` for the whole filing, including the shareholder-fee and expense-example tables that `main` still merged in (`main` produced garbage class names but kept `max_sales_load` and `expense_1yr`/`expense_3yr`).

Relatedly, the flattened single-cell fallback at the end of this function is guarded by the `len(row) > 1` branch, so a flattened cell followed by one empty `<td>` — the shape the third new negative-control test uses — skips the fallback entirely. It only ever fires for rows of exactly length 1.

sambai-dev:
@dgunning Thanks for the detailed repros. I addressed all three review points:

- Prose-only management-fee tables now stop at classification instead of falling through into expense-example, shareholder-fee, or performance tables. Structured mixed tables still retain their real classification.
- Fee-row detection accepts footnote-marked missing values and integer values, while rejecting percent-bearing narrative that does not have a value shape.
- Leading spacer columns are normalized at the table level so multi-class values stay aligned. The same alignment is applied during shareholder-fee extraction, and the flattened-row fallback now remains reachable when trailing cells are empty.

I added end-to-end regressions for the reported first-table-wins losses, marked missing values, single- and multi-class spacer layouts, flattened rows, and indented shareholder fees.

Local verification on the corrected diff:
- 98 focused issue/prospectus tests passed
- 50 characterization tests passed, with the same three environment-dependent metadata cases excluded from that focused command
- fast: 6,433 passed; the 23 failures exactly match the untouched branch
- regression: 2,950 passed; the 10 failures exactly match the untouched branch
- cassette safety, regression provenance, and diff checks passed

PR #1162, opened 2026-08-30
fix(funds): a fund series asked for its filings got the whole trust's (#1143)

### LINKED ISSUE #1143: `FundSeries.get_filings()` returns the whole trust's filings, so a series' newest NPORT-P is a sibling's

dgunning:
Fixes #1143.

## The bug

`FundSeries.get_filings()` delegated straight to the fund company:

```python
return self.fund_company.get_filings(**kwargs)
```

A trust files one report per series under a single CIK, so that returns every sibling series' filings interleaved, and `filings[0]` — the idiom for "the latest one" — belongs to whichever series filed most recently.

```
VEXMX  S000002841  337 filings  top: 0000036405-26-000483
VFINX  S000002839  337 filings  top: 0000036405-26-000483
```

Identical counts and the same top filing for two different funds — and that accession belongs to `S000002846`, a *third* sibling. VEXMX is the Extended Market Index fund, which excludes the S&P 500 by construction, and it was handed the 500 Index fund's holdings.

What makes it bad is that nothing about the answer looks wrong: real filings, from the right trust, that parse cleanly and sum to 100%. Only the series ID inside the document reveals it. Not Vanguard-specific — CIK 734383 behaves the same.

## The fix

`FundSeries` now resolves its own series through SEC browse-edgar — the path `Fund.get_filings(series_only=True)` has used since #888 — and returns an empty `Filings` rather than falling back to the trust, because that fallback is the bug. The resolution logic moved to a module-level `_resolve_series_filings` so both entry points share it rather than duplicating it; most of the diff in `core.py` is that move (byte-identical after dedent). Synthetic `ETF_*` IDs are not SEC series and still use the trust.

```
VEXMX S000002841  n=29  top=...478 -> S000002841
VFINX S000002839  n=28  top=...473 -> S000002839
VGHCX S000002922  n=27  top=...210 -> S000002922
VGENX S000002921  n=27  top=...211 -> S000002921
```

**`get_portfolio()` and `get_latest_report()` were reachable by the same route** and are fixed with it. Both chained through `get_filings()` without `series_only`, so `Fund("VEXMX").get_latest_report()` returned `S000002846`'s report — the same harm through the primary documented holdings entry point, which also backs MCP `edgar_fund(action="portfolio")`. They now request their own series when the `Fund` names one.

## Two things worth a reviewer's attention

**`Fund.get_filings()` without `series_only` is unchanged** and still returns the whole trust. Its default path reached the series entity, so it is now routed to the fund company explicitly — otherwise this fix would have silently made the documented default series-scoped and reduced `series_only=True` to a no-op. The #888 network tests caught that.

**`year`/`quarter` lists were being dropped.** `_series_filter_kwargs` translated only scalars; a list was untranslatable, silently discarded, and the caller received the series' entire unfiltered history — the same silent-wrong-answer shape as the bug being fixed, which this change would have widened from `series_only=True` to the default. Lists are now applied as a union of date ranges, and a pair that genuinely cannot be expressed raises:

```
all: 485   year=2024: 29   year=[2023,2024]: 50   year=2024, quarter=[1,2]: 16
```

## Known limitation, not fixed here

This path returns `Filings`, not `EntityFilings`, so `report_date`, `acceptance_datetime` and `items` are unavailable on the returned filings. Inherent to resolving through browse-edgar, and true of `get_filings(series_only=True)` since #888. Called out in the CHANGELOG rather than left to be discovered; fixing it is separate work.

## Verification

11 regression tests in `tests/issues/regression/test_issue_1143_fund_series_filings.py` (7 offline, 4 network). **All fail on the pre-fix code**, verified by reverting `core.py` and re-running.

The network tests assert the series ID *inside* the returned document — the property the bug violated — rather than counts or accession numbers, both of which drift every quarter (the issue's own numbers had already gone stale). The offline fixture's trust raises if consulted, so those tests pin the *absence of the fallback* rather than the presence of a filter. One test keeps the edgartools-tg7y guarantee that a transport failure raises instead of becoming "no filings."

Full offline suite: **6478 passed, 5 skipped, 2 xfailed, 0 failures.** Lint unchanged from `main` (same 3 pre-existing errors, none in new code).

https://claude.ai/code/session_01LN2NaNcXEuv5YvcKFntcaZ


dgunning:
Reverted in #1165. #1144 by @hashd1ve was already open against #1143 and arrived at substantially the same fix; this PR was written without checking for it and merging it bypassed their work. #1144 lands as the fix for #1143, and the extras this PR carried follow on top of it.

PR #1163, opened 2026-08-30
fix(entity): revenue answered from the wrong year, and the wrong concept (#1149)

### LINKED ISSUE #1149: get_revenue() (and other standardized concept getters) can return a years-stale value after an XBRL tag migration
## Summary

`EntityFacts.get_revenue()` (and any other standardized concept getter built
on `_get_standardized_concept_value`) can return a years-stale value when a
company has migrated its XBRL tag for that concept over time. The stale value
is real data, just from the wrong year, which makes it easy to miss.

## Reproduction

edgartools 5.52.0:

```python

facts = Company("NVDA").get_facts()
print(facts.get_revenue())
```

Actual: `26914000000.0` — NVIDIA's FY2022 revenue.
Expected: `215938000000.0` — NVIDIA's current FY2026 revenue, per its own
FY2026 10-K and matching the "Total Revenue" row of its own income statement
as rendered elsewhere in this library (`edgar_compare`'s embedded
`income_statement` table).

## Cause

`_get_standardized_concept_value` in `edgar/entity/entity_facts.py` tries
each entry in `concept_variants` in a fixed priority order and returns as
soon as **any** variant has **any** matching fact:

```python
for concept in concept_variants:
    for concept_variant in [concept, f'us-gaap:{concept}', f'ifrs-full:{concept}']:
        fact = self.get_annual_fact(concept_variant)
        ...
        if fact and fact.numeric_value is not None:
            unit_result = UnitNormalizer.get_normalized_value(fact=fact, ...)
            if unit_result.success:
                return unit_result.value
```

`get_revenue()`'s `concept_variants` list starts with
`RevenueFromContractWithCustomerExcludingAssessedTax`. NVIDIA tagged revenue
that way through FY2022, then switched to plain `Revenues` from FY2023
onward. `RevenueFromContractWithCustomerExcludingAssessedTax` is tried
first, has real (if 5-year-old) facts, and its most recent one
(`get_annual_fact`) wins immediately — the function never looks at `Revenues`,
which has the current figure, because a higher-priority variant already
"succeeded."

This isn't specific to NVIDIA or to revenue: any company that has migrated
any standardized concept's XBRL tag will have the same problem, and it will
silently return real-looking, wrong-year data rather than erroring.

## Suggested fix

Instead of returning at the first concept variant with any match, gather a
candidate fact from every variant and pick the most recent one across all of
them (by `filing_date`/`period_end`), falling back to an older candidate only
if unit normalization fails for it. Happy to open a PR with this change plus
a regression test if that's welcome.

dgunning:
Fixes #1149. Also fixes internal bead `edgartools-fdye`, which is the same loop.

## Two symptoms, one cause

`_get_standardized_concept_value` walked `concept_variants` in priority order and returned on the first variant with *any* match. That list ranks alternative **names for one figure** — it says nothing about which period a company still tags, or whether a name refers to a total or a slice of one. Ranking by name before anything else produced two distinct wrong answers:

| | contract revenue | `Revenues` | returned |
|---|---|---|---|
| **NVDA** (#1149) | FY2022 26,914,000,000 | FY2026 **215,938,000,000** | the FY2022 value |
| **MET** (fdye) | FY2025 2,436,000,000 | FY2025 **77,084,000,000** | the slice |
| **GIS** (control) | FY2026 **18,424,600,000** | FY2024 2,037,800,000 | correct |

NVIDIA migrated off `RevenueFromContractWithCustomerExcludingAssessedTax`, which stayed top-ranked, so `get_revenue()` answered four-year-old data indefinitely. MetLife tags the ASC-606 slice *and* consolidated `Revenues` for the same year, and priority took the slice — a 32x understatement across insurers and banks, and an income statement showing $2.4B of revenue above $6.1B of operating income.

## The rule

**Period first, then priority.** The newest period any variant reports wins; the priority list then decides among the names that report it. That fixes NVDA without touching priority semantics.

**Then a magnitude cross-check, within that period, opt-in per concept family.** A candidate several times larger than the ranked pick is the consolidated total the pick is a slice of, not a rival name for it. That fixes MetLife.

Two constraints shaped this, and General Mills is why:

- **Magnitude can never be the primary ranking.** Picking the largest outright would prefer `IncludingAssessedTax` over `Excluding`, and gross over net. It is only ever a cross-check on the ranked pick.
- **The cross-check cannot be global.** Net income's variants are not slices of one another — `ProfitLoss` exceeds `NetIncomeLoss` by the noncontrolling interest, and preferring it would quietly change whose earnings are reported. Only `get_revenue()` opts in.

GIS constrains the fix from both sides and is unchanged: its correct value is the *older-ranked* concept in the *newer* period, so "newest wins" alone keeps it right, and its `Revenues` tag is a small slice, so "largest wins" would have broken it.

## The second copy

The same priority list exists again in the statement builder, and MetLife's income statement disagreed with `get_revenue()` for the same company and year. Both now share one threshold (`is_consolidated_total_over` in `entity/utils.py`) rather than two copies that drift — the bead asked for exactly this. The revenue row is also now named after the concept its value came from, instead of the highest-priority concept merely *present*, which had been attributing MetLife's `Revenues` figure to the contract-revenue tag.

## Verification

Measured before and after on live data:

```
NVDA  26,914,000,000  ->  215,938,000,000
MET    2,436,000,000  ->   77,084,000,000
GIS   18,424,600,000  ->   18,424,600,000   (unchanged)
```

Sanity-checked across 13 companies spanning sectors (AAPL, META, JNJ, PRU, AIG, JPM, BAC, XOM, WMT, TSLA) — all revenue, net income and total-asset figures plausible against their own statements; the financials now report consolidated revenue rather than ASC-606 slices.

14 regression tests in `tests/issues/regression/test_issue_1149_stale_standardized_concepts.py` (11 offline on the selection rule, 3 network on the three companies). The offline tests pin the constraints, not just the outcomes: that priority still decides within a period, that the cross-check is off by default, that a near-equal candidate does not override, that the cross-check never crosses periods, and that negatives are not compared by ratio.

Full offline suite: **6488 passed, 5 skipped, 2 xfailed, 0 failures.**

**Note on the mutation check:** the offline tests exercise functions this PR introduces, so on pre-fix code they fail at import rather than on behaviour. The before/after evidence is the measured values above, taken from actual runs on each side.

https://claude.ai/code/session_01LN2NaNcXEuv5YvcKFntcaZ


dgunning:
Closing in favour of #1151 by @biautomator, which was open first against #1149 and which I missed — this PR duplicated it. Apologies for the noise.

Their ranking also carries something this one lacked: a `not is_dimensioned` tiebreak, so a segment-level fact cannot outrank the consolidated one at the same recency. That is a real case I did not handle.

The part of this PR that #1151 does not cover is the `edgartools-fdye` half — the ASC-606 slice that insurers and banks report alongside consolidated `Revenues`. MetLife tags both for FY2025 ($2,436,000,000 and $77,084,000,000), and priority takes the slice, so its income statement shows $2.4B of revenue above $6.1B of operating income. That is being rebuilt on top of #1151 rather than carried here, along with the second copy of the priority list in `enhanced_statement.py` that disagrees with `get_revenue()` for the same company and year.

PR #1214, opened 2026-08-31
fix(xbrl): a Disclosure role about notes payable is not a Notes section (#1207)

### LINKED ISSUE #1207: `XBRL.get_all_statements()` misclassifies disclosure roles containing "notes payable" as Notes
## Issue Type

## Environment
**EdgarTools Version:** 5.53.0  
**Python Version:** 3.14.4  
**Operating System:** Microsoft Windows

## Bug Description

Statement roles whose definitions explicitly identify them as `Disclosure` should be categorized as `disclosure`, even when the subject of the disclosure contains the accounting term "notes payable."

For Oracle's four `Disclosure - NOTES PAYABLE AND OTHER BORROWINGS` roles, `XBRL.get_all_statements()` should report `category="disclosure"`, and the roles should be returned by `xbrl.disclosures()` rather than `xbrl.notes()`.

**What actually happened:**

All four roles are reported as `type="Notes"` and `category="note"`. All four are returned by `xbrl.notes()`, and none is returned by `xbrl.disclosures()`.

The filing has no `FilingSummary` menu category for these roles, so `get_all_statements()` reaches its keyword fallback. The fallback checks the bare substring `"note"` before it checks `"disclosure"`. Because "notes payable" is the disclosure's subject, the first branch wins even though each role definition explicitly begins with the `Disclosure` category marker.

**Error message (if any):**
```text
No exception is raised.

matched definitions: 4
categories: {'note'}
types: {'Notes'}
menu categories: {None}
all returned by xbrl.notes(): True
any returned by xbrl.disclosures(): False
```

## Reproduction
**Company/Ticker:** Oracle Corporation (ORCL)  
**Time Period:** Q1 FY2024, period ended August 31, 2023  
**Relevant Forms:** 10-Q accession `0000950170-23-047713`

```python
from pathlib import Path

from edgar.xbrl import XBRL

# Run from a directory containing the unpacked accession package.
xbrl = XBRL.from_directory(
    Path("0000950170-23-047713") / "xbrl"
)

rows = [
    row
    for row in xbrl.get_all_statements()
    if "DISCLOSURE - NOTES PAYABLE AND OTHER BORROWINGS"
    in str(row.get("definition", "")).upper()
]

note_roles = {statement.role_or_type for statement in xbrl.notes()}
disclosure_roles = {
    statement.role_or_type for statement in xbrl.disclosures()
}
target_roles = {row["role"] for row in rows}

print("matched definitions:", len(rows))
print("categories:", {row["category"] for row in rows})
print("types:", {row["type"] for row in rows})
print("menu categories:", {row["menu_category"] for row in rows})
print(
    "all returned by xbrl.notes():",
    target_roles <= note_roles,
)
print(
    "any returned by xbrl.disclosures():",
    bool(target_roles & disclosure_roles),
)
```

## Additional Context

The four affected role definitions are the base disclosure plus its Tables, Details, and Narrative Details roles. Their primary concepts are `us-gaap_DebtDisclosureAbstract` or `us-gaap_DebtInstrumentsAbstract`; none is `us-gaap_NotesToFinancialStatementsAbstract`.

At the pinned 5.53.0 source commit, the [`get_all_statements()` keyword fallback](https://github.com/dgunning/edgartools/blob/aeb15ba5562e5bdb707fa99d201d35a5f3ff600e/edgar/xbrl/xbrl.py#L938-L945) checks:

```python
if 'us-gaap_NotesToFinancialStatementsAbstract' in primary_concept or 'note' in role_def:
    statement_type = "Notes"
    statement_category = "note"
elif 'us-gaap_DisclosuresAbstract' in primary_concept or 'disclosure' in role_def:
    statement_type = "Disclosures"
    statement_category = "disclosure"
```

Both keywords occur in each Oracle role definition, but the unanchored `note` test matches the business term "notes payable" first. A category-prefix-aware rule, or another precedence rule that distinguishes a `Disclosure` marker from words in the disclosure title, would avoid this false positive. A regression fixture named `Disclosure - Notes Payable` would cover the collision directly.

This is a role-classification defect, not a missing-content or wrong-value claim. The four statements remain retrievable, but they appear on the wrong category route. It is distinct from [Third Wave draft 30](../Feature%20Request/30-xbrl-notes-disclosure-routing.md), which requests clearer documentation and an opt-in unified route even when statement categories are classified correctly.

A bounded query of the local EdgarTools issue catalog for statement-category, Notes, disclosure, and keyword-classification terms did not identify an exact match. The open and closed GitHub tracker should be rechecked immediately before submission.

---

manantlerio:
Fixes #1207.

## The defect

`get_all_statements()` falls back to keyword matching when a role carries no `FilingSummary` menu category, and that fallback tested the bare substring `"note"` before it tested `"disclosure"`:

```python
if 'us-gaap_NotesToFinancialStatementsAbstract' in primary_concept or 'note' in role_def:
    statement_type = "Notes"
elif 'us-gaap_DisclosuresAbstract' in primary_concept or 'disclosure' in role_def:
    statement_type = "Disclosures"
```

"Note" names a financial-statement section and a debt instrument, and both keywords are present in a definition like `0000021 - Disclosure - NOTES PAYABLE AND OTHER BORROWINGS`. The first branch won, so Oracle's four notes-payable roles (10-Q `0000950170-23-047713`) reported `type="Notes"` / `category="note"`, came back from `xbrl.notes()`, and were absent from `xbrl.disclosures()`.

## The change

The keyword order is unchanged. What changed is that the ambiguous word no longer outranks a role that states what it is:

- its `Disclosure` category marker - either the `" - "` segment of an EDGAR role definition, or the marker leading the role name when the schema definition is missing (`DisclosurePromissoryNotesPayable`); or
- the concept it hangs from, `us-gaap_DebtDisclosureAbstract`. `Statements.classify_statement` (tier 2) and the `Disclosures` concept patterns in `statement_resolver.py` already read a `*DisclosureAbstract` concept as a disclosure; this fallback tested only the literal `us-gaap_DisclosuresAbstract`, which these roles never carry.

unless the definition names the notes section itself - `Notes to Consolidated Financial Statements`, `Note 1 - Organization`, `Footnotes`, a role titled only `Notes`. That test is deliberately generous: a definition matching it keeps the classification it already had, so a false positive costs nothing while a false negative would move a real note.

**A role whose definition does not contain "note" takes exactly the branch it took before.** The change is confined to the disputed branch.

## Verification

`tests/issues/regression/test_issue_1207_disclosure_notes_payable.py`. Two of the fixtures already committed to the repository carry the defect, so the end-to-end assertions run offline with no cassette:

| fixture | role | before | after |
|---|---|---|---|
| `datafiles/gahc` | `PromissoryNotesPayable` (`us-gaap_DebtDisclosureAbstract`) | `note` / `Notes` | `disclosure` / `Disclosures` |
| `datafiles/gahc` | `ConvertiblePromissoryNotesPayable` | `note` / `Notes` | `disclosure` / `Disclosures` |
| `datafiles/aeon` | `DisclosureDaewoongConvertibleNotesPredecessor` | `note` / `Notes` | `disclosure` / `Disclosures` |
| `datafiles/aapl` | no definition contains "note" | unchanged | unchanged |

On `main` those assertions fail 6 ways; with the fix they pass. The unit-level cases cover the four Oracle definitions from the report verbatim, plus the notes sections that must not move.

Also run, on Windows / Python 3.14: the 864 non-network tests matching `xbrl or statement or notes or disclosure or categor` (848 pass; the 16 failures in `test_notes_characterization.py` are a `cp1252` decode of the baseline files and fail identically on `main`), and `scripts/check_regression_provenance.py` / `scripts/check_regression_skips.py`, both OK. `ruff check edgar/xbrl/xbrl.py` reports the same 15 pre-existing findings before and after.

## Behaviour note

For a filing with no `FilingSummary`, a notes-payable role now leaves `xbrl.notes()` and appears in `xbrl.disclosures()`, which is what the issue asks for. It is still reachable by role and by name - `xbrl.get_table("...")` and `xbrl.get_disclosure(role)` are not category-filtered. Filings that do have a `FilingSummary` never reach this fallback and are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A disclosure whose subject is notes payable was routed to `notes()`.** `get_all_statements()` falls back to keyword matching when a role carries no `FilingSummary` menu category, and that fallback tested the bare substring `"note"` before it tested `"disclosure"`. "Note" names a financial-statement section and a debt instrument, so Oracle's four `Disclosure - NOTES PAYABLE AND OTHER BORROWINGS` roles (10-Q `0000950170-23-047713`) reported `type="Notes"` / `category="note"`, came back from `xbrl.notes()`, and were absent from `xbrl.disclosures()`. The ambiguous word no longer outranks a role that states what it is - by its `Disclosure` category marker, or by the concept it hangs from (`us-gaap_DebtDisclosureAbstract`) - unless the definition names the notes section itself, so `Notes to Consolidated Financial Statements` and `Note 1 - Organization` still classify as notes. Roles whose definition does not contain "note" are classified exactly as before. (GH #1207)
+
 ## [5.55.0] - 2026-08-31
 
 ### Changed
--- a/edgar/xbrl/xbrl.py
+++ b/edgar/xbrl/xbrl.py
@@ -1072,6 +1072,18 @@ class XBRL:
                     # notes payable, so a role that declares itself a disclosure
                     # is taken at its word unless the definition names the notes
                     # section itself (issue #1207).
+                    #
+                    # Known limitation (issue #1218): this decides one role at a
+                    # time, and in a filing that falls back to role names the
+                    # members of one family do not all carry the same evidence.
+                    # gahc's `ConvertiblePromissoryNotesPayable` hangs from
+                    # us-gaap_DebtDisclosureAbstract and moves; its `Tables` and
+                    # `ScheduleOf...` children hang from a debt-balance abstract
+                    # and their names do not lead with the marker, so they stay
+                    # notes and the family splits across the two accessors. A
+                    # role with a real schema definition - the case in #1207 -
+                    # carries ` - Disclosure - ` on every member, so those move
+                    # together.
                     if _names_notes_section(role_def) or not _declares_disclosure(role_def, primary_concept):
                         statement_type = "Notes"
                         statement_category = "note"
--- a/edgar/xbrl/xbrl.py
+++ b/edgar/xbrl/xbrl.py
@@ -12,6 +12,7 @@ organizing facts according to presentation hierarchies, validating calculations,
 and handling dimensional qualifiers.
 """
 import datetime
+import re
 from pathlib import Path
 from textwrap import dedent
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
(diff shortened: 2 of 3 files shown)


dgunning (commented):
Reviewed by verifying the claims independently rather than reading the reasoning. **The fix is real, correctly scoped, and the tests genuinely gate it.** One finding below, which I'd call a follow-up rather than a blocker.

## What I checked

**The load-bearing assumption holds.** Both helpers document `role_def` as lowercased, and `edgar/xbrl/xbrl.py:961` is `role_def = tree.definition.lower()`. `primary_concept` is *not* lowercased, which is why `_DISCLOSURE_CONCEPT_RE` carries `IGNORECASE` — correct.

**The mutation claim is true.** I couldn't run your test file against `main` directly (see the note at the end), so I extracted the three end-to-end tests into a standalone module without the helper imports:

- against this branch: 3 passed
- against `main`'s `xbrl.py`: 2 failed with exactly `AssertionError: assert 'note' == 'disclosure'`, and `test_aapl_classification_is_unchanged` passed

So the control passes either way and the two real cases fail without the fix. That is what a gate looks like.

**The blast radius is what you say it is.** I classified every role in all 8 committed fixture filings, before and after:

```
total roles: 347   moved: 8
```

All 8 are genuine notes-payable / convertible-notes disclosures (5 aeon, 3 gahc). Nothing else moved — no collateral damage across msft, nflx, tsla, unp, aes, aapl.

**`hatch run test-fast` on this branch: 6600 passed, 9 skipped, 1 xfailed.** No failures. (6579 on the `2e33a123` base + your 21.)

**Fixture discipline is right.** All three fixture directories are tracked (6/6, 7/7, 6/6 files), none gitignored, and the tests `assert directory.exists()` rather than `pytest.skip`. A missing fixture fails loudly instead of quietly reporting green — this repo has been bitten by the other pattern.

## Finding: a role family now splits across `notes()` and `disclosures()`

The 4 note-bearing roles that did *not* move are the same subject as ones that did, in the same filing:

| gahc role | primary concept | result |
|---|---|---|
| `ConvertiblePromissoryNotesPayable` | `us-gaap_DebtDisclosureAbstract` | → **disclosure** |
| `ConvertiblePromissoryNotesPayableTables` | `us-gaap_LongtermConvertibleDebtCurrentAndNoncurrentAbstract` | stays **note** |
| `ConvertiblePromissoryNotePayableScheduleOf…` | same | stays **note** |
| `ConvertiblePromissoryNotesPayableScheduleOfRollfoward…` | same | stays **note** |

`_declares_disclosure` returns `False` for the children: their role *names* don't lead with `Disclosure`, and their concept is a debt-balance abstract rather than a `*DisclosureAbstract`.

This doesn't affect the Oracle case in the issue — those roles have real schema definitions carrying ` - Disclosure - `, so all four move together. It only bites filings that fall back to role names, which is exactly where gahc sits.

Before this change those four were consistently (if wrongly) `note`. After it, the parent narrative is in `disclosures()` while its Tables and Schedules stay in `notes()`, so someone iterating either accessor gets a partial family. That's a smaller problem than the one you're fixing, and I don't think it should hold the PR — but it's worth a note in the code or a follow-up issue, because "the parent moved and its details didn't" is the kind of thing that gets rediscovered as a new bug later.

## Minor: the test module can't be mutation-checked as written

`from edgar.xbrl.xbrl import XBRL, _declares_disclosure, _names_notes_section` at module level means reverting `xbrl.py` turns the whole file into a collection error rather than failures. That's how I ended up extracting the end-to-end tests by hand to verify your claim. Not worth changing on its own, but importing the private helpers lazily inside the unit tests would let the next person run the file against an unfixed tree and see real failures.

## Nits

- `_declares_disclosure` matches the ` - ` separated segment or a leading marker. A definition using an en-dash, or no spaces around the hyphen, falls through to the old behaviour. That's the safe direction and matches your stated design, so this is a note rather than a request.
- `_NOTES_SECTION_RE`'s `notes?\s*\d` will match a definition like `Note 7 - Income Taxes`, which is right. I checked it does *not* match `notes payable`, since no digit follows.

Nice piece of work — the reasoning about false positives costing nothing while false negatives move a real note is the right way round, and the PR body made it fast to check.

Issue #459, opened 2025-10-11
Old Filings (with no xbrl format) are  not convertable to dataframe


jaydjunior:
## Issue Type

- [ ] Data Quality Issue (incorrect financial values, missing data)
- [x] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** (4.18.0)
**Python Version:** (3.13)  
**Operating System:** (Windows 11)

## Bug Description
**What you expected to happen:**
I expected to be able to convert multiple balance sheets of one ticker to a dataframe. My program crashes with an error.

**What actually happened:**
```
File "...\a.py", line 20, in <module>
    income_statements = stitched_statements.income_statement().to_dataframe()
  File "...\.venv\Lib\site-packages\edgar\xbrl\statements.py", line 1011, in to_dataframe
    return to_pandas(self.statement_data)
                     ^^^^^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\edgar\xbrl\statements.py", line 972, in statement_data
    self._statement_data = self.xbrls.get_statement(
                           ~~~~~~~~~~~~~~~~~~~~~~~~^
        self.statement_type,
        ^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
        self.include_dimensions
        ^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "...\.venv\Lib\site-packages\edgar\xbrl\stitching\xbrls.py", line 164, in get_statement
    result = stitch_statements(
        self.xbrl_list,
    ...<5 lines>...
        include_dimensions=include_dimensions
    )
  File "...\.venv\Lib\site-packages\edgar\xbrl\stitching\core.py", line 557, in stitch_statements
    optimal_periods = determine_optimal_periods(xbrl_list, statement_type, max_periods=max_periods)
  File "...\.venv\Lib\site-packages\edgar\xbrl\stitching\periods.py", line 543, in determine_optimal_periods
    return optimizer.determine_optimal_periods(xbrl_list, statement_type, max_periods)
[...]

Issue #1174, opened 2026-08-30
`XBRLS.from_filings()` silently swallows unexpected parsing exceptions and returns partial results


synfonia-llc:
## Issue Type

- [ ] Data Quality Issue (incorrect financial values, missing data)
- [x] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.53.0
**Python Version:** 3.14.4  
**Operating System:** Microsoft Windows 11 Pro 10.0.26200

## Bug Description
**What you expected to happen:**
An unexpected exception raised while parsing an XBRL filing should produce an observable failure signal that identifies the affected filing, rather than being silently converted into an apparently successful partial `XBRLS` result. That signal could be an exception with filing context, structured failure metadata, or an explicit warning/strict-mode contract; the important behavior is that the caller can detect which requested filing failed and why.

Intentional tolerance for filings that simply have no XBRL data should remain unchanged. A fix can preserve that behavior by explicitly skipping `XBRL.from_filing()` results of `None` while making unexpected exceptions observable.

**What actually happened:**
`XBRLS.from_filings()` catches every `Exception` without logging, recording, or re-raising it. The method returns normally with only the successfully parsed filing, so callers cannot tell that one requested filing was lost. The reproduction prints:

```
returned normally
parsed XBRL count: 1
parsed entities: ['Good Co']
```

**Error message (if any):**
```
No error is emitted. RuntimeError("parser invariant failed") is silently swallowed.
```

## Reproduction
**Company/Ticker:** N/A (generic exception-handling path; deterministic fault injection is required to prove it reliably)
**Time Period:** Synthetic FY 2024-2025
**Relevant Forms:** 10-K (synthetic filing metadata)

**Minimal code to reproduce:**
```python
from datetime import date
from types import SimpleNamespace
from unittest.mock import patch

from edgar.xbrl import XBRLS

filings = [
    SimpleNamespace(
        accession_no="good", form="10-K", filing_date=date(2025, 2, 1)
    ),
    SimpleNamespace(
        accession_no="broken", form="10-K", filing_date=date(2024, 2, 1)
    ),
]


def parse(filing):
    if filing.accession_no == "broken":
[...]

PR #1313, opened 2026-09-08
fix(xbrl): a filing that fails to parse during stitching disappears with no trace

AmirF194:
`XBRLS.from_filings()` wraps `XBRL.from_filing(filing)` in `except Exception: pass`, so any unexpected parser failure disappears with no trace of which filing it was or why. The caller gets a normally-returned `XBRLS` built from only the filings that happened to succeed, indistinguishable from a fully successful multi-filing parse.

I reproduced this with the fault-injection repro from the issue (mocked `XBRL.from_filing` raising on one of two filings): the result comes back with one entry and no error, warning, or log line pointing at the one that failed.

The fix logs the accession number, form and exception for the filing that failed, so the caller (or anyone reading logs from a batch job) can tell which filing was lost and why. It leaves the pre-existing, intentional tolerance for a filing with no XBRL data at all untouched: `XBRL.from_filing()` returning `None` for that case (issue #459) never raises, so it still reaches `xbrl_list` unchanged and produces no log line.

## Verification

- New regression test (`tests/issues/regression/test_issue_1174_xbrls_swallowed_exceptions.py`, 3 cases): fails on `main` (no signal is produced for the failing filing), passes on this branch, confirmed both ways in the same container.
- A fourth case confirms the `#459` no-XBRL tolerance still produces no log line and still reaches `xbrl_list`.
- `ruff check edgar/xbrl/stitching/xbrls.py` and `scripts/check_cassettes.py` are clean; this change touches no cassettes.
- Not run: the full network/slow suites (this fix touches no network path, and the new tests are fully offline/mocked).

### DIFF
--- a/edgar/xbrl/stitching/xbrls.py
+++ b/edgar/xbrl/stitching/xbrls.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 import pandas as pd
 
+from edgar.core import log
 from edgar.xbrl.stitching.core import StatementStitcher, stitch_statements
 from edgar.xbrl.stitching.query import StitchedFactQuery, StitchedFactsView
 
@@ -80,8 +81,12 @@ class XBRLS:
             try:
                 xbrl = XBRL.from_filing(filing)
                 xbrl_list.append(xbrl)
-            except Exception:
-                pass
+            except Exception as exc:
+                log.warning(
+                    f"XBRLS.from_filings: filing {getattr(filing, 'accession_no', filing)} "
+                    f"({getattr(filing, 'form', '?')}) raised {exc.__class__.__name__}: {exc}, "
+                    f"skipped from the result."
+                )
 
         return cls(xbrl_list)
 
--- a/edgar/xbrl/stitching/xbrls.py
+++ b/edgar/xbrl/stitching/xbrls.py
@@ -80,12 +80,20 @@ class XBRLS:
         for filing in sorted_filings:
             try:
                 xbrl = XBRL.from_filing(filing)
+                if xbrl is None:
+                    log.debug(
+                        f"XBRLS.from_filings: filing {getattr(filing, 'accession_no', filing)} "
+                        f"({getattr(filing, 'form', '?')}) has no XBRL attachments, "
+                        f"skipped from the result."
+                    )
+                    continue
                 xbrl_list.append(xbrl)
             except Exception as exc:
                 log.warning(
                     f"XBRLS.from_filings: filing {getattr(filing, 'accession_no', filing)} "
                     f"({getattr(filing, 'form', '?')}) raised {exc.__class__.__name__}: {exc}, "
-                    f"skipped from the result."
+                    f"skipped from the result.",
+                    exc_info=True,
                 )
 
         return cls(xbrl_list)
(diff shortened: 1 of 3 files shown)


edgar/xbrl/stitching/xbrls.py
```diff
@@ -80,8 +81,12 @@ def from_filings(cls, filings: Union['Filings', List[Any]], filter_amendments:bo
             try:
                 xbrl = XBRL.from_filing(filing)
                 xbrl_list.append(xbrl)
```
dgunning, line 90: The raising path is now covered, but the `None` path still poisons `xbrl_list`, and that's the half issue #1174 explicitly asked for ("explicitly skip `XBRL.from_filing()` results of `None`").

`XBRL.from_filing()` returns `None` for a filing with no XBRL attachments (`edgar/xbrl/xbrl.py:829`, the `xbrl_attachments.empty` branch) and that `None` gets appended here. When the newest filing is the one without XBRL, `XBRLS.from_filings(...)` raises `AttributeError: 'NoneType' object has no attribute 'entity_info'` straight out of `XBRLS.__init__` (`xbrls.py:40`). When the `None` isn't first it survives into `xbrl_list` and detonates later instead, at `edgar/xbrl/statements.py:3742` and `:3770` where `xbrl.get_all_statements()` is called on it.

The issue you're describing in the docstring — that `None` is a true absence and not a failure — is right, but the consequence is that it should be skipped rather than carried. Suggest:

```python
xbrl = XBRL.from_filing(filing)
if xbrl is not None:
    xbrl_list.append(xbrl)
else:
    log.debug(...)
```

The new `test_none_for_no_xbrl_is_not_treated_as_a_failure` case asserts `result.xbrl_list[1] is None`, which pins the current broken behaviour as the contract — worth relaxing that to assert only that no warning was logged, so the real fix doesn't have to come back as a test change.

dgunning (commented):
Thanks for this — the diagnosis is right and the logging is the correct shape. Two things before merge, both in `from_filings`. The changelog placement is on us, not you: the fragment convention in `changelog.d/` wasn't advertised anywhere you'd have seen it, so leave `CHANGELOG.md` as is and we'll sort it out on our side.

edgar/xbrl/stitching/xbrls.py
```diff
@@ -80,8 +81,12 @@ def from_filings(cls, filings: Union['Filings', List[Any]], filter_amendments:bo
...
-            except Exception:
-                pass
+            except Exception as exc:
+                log.warning(
```
dgunning, line 92: Worth adding `exc_info=True` here. The formatted message carries the exception class and `str(exc)`, which identifies the filing but not the failure site — for an internal parser invariant blowing up (a `KeyError: 'segment'`, an `IndexError` deep in `parse_instance_content`) an operator reading batch logs learns which filing died and nothing about where. This branch only runs on failure, so the traceback is free.

edgar/xbrl/stitching/xbrls.py
AmirF194: Fixed, skips the None case now instead of appending it, and relaxed the pinned test to assert on the count rather than xbrl_list[1] is None.

edgar/xbrl/stitching/xbrls.py
AmirF194: Added exc_info=True, good catch.

PR #1319, opened 2026-09-11
fix: surface ambiguous 13F value units and support explicit overrides.

JonArnfred:
Warn when an uncertain unit inference would multiply values by 1,000. Expose raw holdings and unit-resolution diagnostics, and allow per-filing dollar or thousand-dollar overrides across holdings and summary totals.

Preserve automatic unit choices and bypass accession-only caches for overrides. Add original Kahn Brothers Q1 2022 fixtures, regression tests, documentation, and a changelog fragment


dgunning:
Thanks for this change, it measurable improves the error handling of the feature

PR #1329, opened 2026-09-16
fix(xbrl): a disclosure's Tables and Details follow their parent when a filing falls back to role names

### LINKED ISSUE #1218: Role family splits across notes() and disclosures() when a filing falls back to role names (follow-up to #1207)
## Issue Type

## Environment
**EdgarTools Version:** `main` (post #1214)
**Python Version:** 3.14
**Operating System:** Windows

## Bug Description

Follow-up to #1207 / #1214, raised by @dgunning in review of that PR. Filed so it is
tracked rather than rediscovered later as a new bug.

**What happens:** in a filing whose roles fall back to role *names* (no human-readable
schema definition), a disclosure's parent role can be classified `disclosure` while its
`Tables` and `Schedule...` children stay `note`. Iterating either `xbrl.notes()` or
`xbrl.disclosures()` then returns a partial family.

**What you would expect:** a role family classifies as a unit. The Tables and Details
belonging to a disclosure follow the disclosure.

## Reproduction

`data/xbrl/datafiles/gahc` (Global Arena Holding 10-Q), committed to the repository, so
this reproduces offline:

```python
from pathlib import Path
from edgar.xbrl import XBRL

xbrl = XBRL.from_directory(Path("data/xbrl/datafiles/gahc"))
for s in xbrl.get_all_statements():
    if "ConvertiblePromissoryNote" in s["definition"]:
        print(f'{s["category"]:11} {s["primary_concept"]:60} {s["definition"]}')
```

| role | primary concept | category |
|---|---|---|
| `ConvertiblePromissoryNotesPayable` | `us-gaap_DebtDisclosureAbstract` | **disclosure** |
| `ConvertiblePromissoryNotesPayableTables` | `us-gaap_LongtermConvertibleDebtCurrentAndNoncurrentAbstract` | note |
| `ConvertiblePromissoryNotePayableScheduleOfConvertiblePromissoryNotesPayableDetails` | same | note |
| `ConvertiblePromissoryNotePayableScheduleOfConvertiblePromissoryNotesPayableDetailsParenthetical` | same | note |
| `ConvertiblePromissoryNotesPayableScheduleOfRollfowardOfConvertiblePromissoryNotesPayableDetails` | same | note |

## Cause

`_declares_disclosure()` in `edgar/xbrl/xbrl.py` returns `False` for the children. Neither
of its two signals is present on them:

- their role *names* do not lead with the `Disclosure` marker (`ConvertiblePromissoryNotesPayableTables`), and
- their primary concept is a debt-balance abstract (`us-gaap_LongtermConvertibleDebtCurrentAndNoncurrentAbstract`) rather than a `*DisclosureAbstract`.

The parent carries `us-gaap_DebtDisclosureAbstract` and so moves on its own.

## Scope

This does not affect the case in #1207. Oracle's roles have real schema definitions
carrying ` - Disclosure - `, so the base role and its Tables/Details/Narrative all move
together. It bites only filings that fall back to role names, which is where `gahc` sits.

Before #1214 those four children were consistently, if wrongly, `note`; they are not newly
wrong, they are newly *inconsistent with their parent*.

## Possible directions

Not proposing one - noting what seems available:

1. **Propagate from the parent.** Group roles by a normalised name stem (strip a trailing
   `Tables` / `Details` / `Policies` / `ScheduleOf...`) and let a family member that
   declares itself a disclosure carry the rest. Closest to how a filer means it, and the
   most likely to have edge cases.
2. **Widen `_declares_disclosure` for children.** Recognise the `Tables`/`Details` suffix
   and look up the stem's classification. A narrower version of 1.
3. **Leave classification alone and fix it at the accessor.** Have `notes()` and
   `disclosures()` return whole families.

Whichever way, the guard is the same measurement #1214 used: classify every role in all 8
committed fixture filings before and after and confirm only the intended ones move.

wolfgang-aura:
Closes #1218.

In a filing whose roles fall back to role *names*, a disclosure's `Tables` and `Details` roles were classified `note` while their parent was `disclosure`, so `xbrl.disclosures()` returned the parent alone and `xbrl.notes()` returned the four children. The committed `data/xbrl/datafiles/gahc` 10-Q shows it: `ConvertiblePromissoryNotesPayable` hangs from `us-gaap_DebtDisclosureAbstract` and moves, its `...Tables` and `...ScheduleOf...Details` children hang from a debt-balance abstract and their names do not lead with the marker, so `_declares_disclosure()` is false for them and the family splits.

The fix is direction 2 from the issue, kept to the ambiguous `'note' in role_def` branch of `get_all_statements()`. A pre-pass over the presentation trees collects a family key for every role on that path and records which stems declare a disclosure. A role ending in a family suffix (`Tables`, `Details`, `Policies`, `Parenthetical`, `Textuals`) then resolves to the *longest* stem whose CamelCase segments lead its own, segment by segment, and follows it only if that stem declares. Longest-match matters: with both `NotesPayable` (declares) and `NotesPayableRelatedParty` (does not), `NotesPayableRelatedPartyDetails` stays with its own stem instead of jumping to the shorter one. A segment matches its plural (`Note` / `Notes`) because gahc spells the parent `...NotesPayable` and two children `...NotePayableScheduleOf...`; a raw character prefix is never used, so `Debt` does not claim `DebtorNotesDetails`. `_declares_disclosure()` and the `_names_notes_section()` guard from #1214 are unchanged; roles with a `FilingSummary` category or a ` - Disclosure - ` schema definition never reach the new code. About 270 lines of source, 130 of tests, two changelog fragments.

### How this was tested

Fixtures used: the seven committed `data/xbrl/datafiles` filings (aapl, aeon, gahc, msft, nflx, tsla, unp), offline. Classifying all 347 roles before and after: exactly the four gahc `ConvertiblePromissoryNote...` children move `note -> disclosure`; nothing else changes.

New tests live in `tests/issues/regression/test_issue_1207_disclosure_notes_payable.py` under a `#1218` section, since this is the follow-up on the same fixture: parametrized positive cases, negative cases (a stem, a foreign family's child, a suffix-less sibling, the `NotesPayable` / `NotesPayableRelatedParty` / `...Details` triple), and an end-to-end gahc check of category, type and both accessors.

- `pytest tests/issues/regression/test_issue_1207_disclosure_notes_payable.py tests/xbrl/test_statement_drilldown.py tests/company_reports/test_notes.py -q -m "not network"` on the revised tree: 136 passed (the regression file alone: 34).
- `tests/xbrl` without network tests: 870 passed; the 22 failures and 12 errors in `test_notes_characterization.py`, `test_xbrl_facts.py` and `test_xbrl_statements.py` are identical at the base commit `35e3c324` in this environment, so they predate this branch.

Run on Windows 11 / Python 3.14.3 with pyarrow 24.0.0 and lxml 6.1.1; CI covers the rest of the matrix.

### Revision after review

Answers to the four points in the review of 2026-09-20:

1. **XBRL-only Notes builder.** `Notes._build_from_xbrl_only` now takes `note`- and `disclosure`-category stems and hangs their `Tables`, `Policies` and `Details` members from them by name, using the same family helpers as the classifier. Measured with `XBRL.from_directory` (no FilingSummary): gahc `Notes.from_xbrl` goes 0 -> 7 on this branch, aapl 0 -> 16, and `Convertible Promissory Notes Payable` carries 1 table, 3 details, 0 policies, so `get_notes_for_concept('us-gaap_ConvertibleNotesPayableCurrent', xbrl)` returns it. `test_gahc_notes_are_still_built_without_a_filing_summary` pins all of that; the `== {"AccountingPolicies"}` assertion stays because it is now true for the right reason rather than load-bearing.
2. **Stem match on a segment boundary.** `_role_family_key` returns lowercased CamelCase segments and `_is_family_stem_of` compares them pairwise. `test_a_stem_claims_only_names_that_extend_it_on_a_segment_boundary` carries both probes from the review: `Debt` claims `DebtDetails`, not `DebtorNotesDetails`; `ConvertibleNote` claims `ConvertibleNotesDetails`, not `ConvertibleNoteholderRightsDetails`.
3. **Empty family key.** `_role_family_stem` skips a stem with no segments, so a role URI ending in `/` claims nothing.
4. **`-ies` / `-y` plurals.** Deliberately not handled; `_segments_match` tolerates a trailing `s` only and its docstring now says so, with `Inventory` / `InventoriesDetails` as the shape that stays unmatched. No committed fixture exhibits it; widening the rule is a one-line change when one does.

The `_CAMEL_PLURAL_RE` collisions (`BasisOfPresentation` / `AssetsHeld`) disappear with the character-level stripping.

### An alternative I did not take

Direction 3, fixing it at the accessor so `notes()` and `disclosures()` return whole families, would have left `get_all_statements()` reporting the split the issue's table shows, so callers reading `category` directly would still see it. Direction 1, propagating from the parent without a suffix check, would let any declared stem claim any longer role name; the suffix list bounds what can follow. One visible trade-off: `Policies` is in that list, so a hypothetical `NotesPayablePolicies` under a declaring stem now classifies `disclosure`; no fixture has one.

### AI disclosure

This change was drafted with AI assistance: Claude Opus 5 wrote the patch and a second Claude Opus 5 pass reviewed it, both running under Mailman. The test results above come from the harness executing the commands itself, not from either agent's account of its own work. The account filing this pull request answers for every line of it.


dgunning (changes requested):
Thank you for this — the diagnosis is right, the restructuring of the `'note' in role_def` branch is behaviour-preserving for the old paths, and the negative unit tests (especially the `NotesPayable` / `NotesPayableRelatedParty` triple) are the right instinct. I verified your headline claim and it holds exactly: 347 roles across the committed fixtures, exactly the four gahc `ConvertiblePromissoryNote...` children move, nothing else changes. The AI disclosure is appreciated, and the harness-reported numbers checked out.

There is one blocking problem, and it is downstream of the classification rather than in it.

## Blocker: the XBRL-only Notes builder goes empty

`Notes._build_from_xbrl_only` (`edgar/xbrl/notes.py:463`) selects roles where `category == 'note' and type == 'Notes'`. #1207 already moved the family stem to `disclosure`; this PR moves the four children, so that bucket is now empty for exactly the filings this change targets.

Measured on the gahc fixture:

```
main:   Notes.from_xbrl(xbrl) -> 4     ('note','Notes'): 4, ('disclosure','Disclosures'): 3
branch: Notes.from_xbrl(xbrl) -> 0     ('note','Notes'): absent, ('disclosure','Disclosures'): 7
```

Because `_get_concept_to_notes_index` (`notes.py:1245`) is built from that list, `StatementLineItem.note`, `.notes` and `get_notes_for_concept()` return nothing for every line item in such a filing. This path runs whenever `xbrl._filing_summary` is `None`, which includes every `XBRL.from_directory(...)` caller and any filing whose FilingSummary is unavailable — gahc itself has `_filing_summary is None`.

The end-to-end test currently encodes this rather than catching it:

```python
assert {s["type"] for s in xbrl.get_all_statements() if s["category"] == "note"} == {"AccountingPolicies"}
```

That passes precisely because the family left the note bucket. The fallback builder probably needs to accept `disclosure`-category roles while skipping the Tables/Details children, which `_ROLE_FAMILY_SUFFIX_RE` now makes identifiable.

## Precision of the stem match

`family_key.startswith(key)` is a raw character prefix with no CamelCase boundary, so a short declaring stem claims longer unrelated names. Within the `'note'` guard that bounds the blast radius, but not enough — short generic stems are exactly what filers use in role-name fallback. Verified against the shipped helpers:

```
stem Debt (declares)            -> DebtorNotesDetails                  FOLLOWS
stem ConvertibleNote (declares) -> ConvertibleNoteholderRightsDetails  FOLLOWS
```

Neither is a member of the stem's family. Matching on a CamelCase segment boundary rather than a character prefix closes this.

Two related edges:

- `_CAMEL_PLURAL_RE` strips any `s` before a capital or end of string, plural or not: `BasisOfPresentation` -> `basiofpresentation`, `AssetsHeld` -> `assetheld`, `GrossRevenue` -> `grosrevenue`. Combined with `stem_declares_disclosure[key] = ... or declares`, two genuinely different stems that collide after stripping share one declaring flag.
- An empty key claims everything. `presentation.py:86-87` falls back to `role_id = role.split('/')[-1]`, which is `''` for a role URI ending in `/`; `_role_family_key('')` is `''`, and `family_key.startswith('')` is true for every role. Confirmed: `_follows_disclosure_family('convertiblenotesdetails', 'convertiblenotedetail', {'': True})` returns `True`. A `if not key: continue` in the pre-pass removes it.

## On the corpus evidence

Worth knowing for the next iteration: the before/after comparison is weaker than it reads. Instrumenting `_follows_disclosure_family`, only 4 of the 347 roles reach it at all, and all four return `True`. No role in the corpus reaches the rule and is correctly rejected by it, so "nothing else changes" is guaranteed by construction rather than evidence of precision. Three of the eight `data/xbrl/datafiles` directories also yield zero roles (`aes` is a lone HTML file, `msft` has only the instance and schema, `nflx` has year subdirectories), so the effective corpus is five filings, one of which exercises this code.

Separately, the plural handling only covers a trailing `s`, not `-y`/`-ies`: `Inventory` + `InventoriesDetails`, `PropertyAndEquipment` + `PropertiesAndEquipmentDetails` and `Liability` + `LiabilitiesDetails` all fail to match. That is the same singular/plural shape gahc exhibits, so it is a gap in the fix rather than a defect.

## What I would change

1. Fix the XBRL-only Notes builder so the family remains reachable through `StatementLineItem.note` and `get_notes_for_concept()`, and revise the end-to-end assertion so it no longer locks in the empty result.
2. Match the stem on a CamelCase segment boundary rather than a raw prefix.
3. Skip empty family keys in the pre-pass.
4. Optionally extend the key normalisation to `-ies`/`-y`, or say in the comment that it is deliberately limited to trailing `s`.

The core of this is good work and I would like to take it once the Notes path is sound. Happy to look at the next revision quickly.

dgunning (approved):
Re-reviewed against `0fe9fa09`. All four points are addressed, I verified each one rather than taking the commit message for it, and the blocker turned out to be hiding something bigger. Approving.

## The blocker is fixed — and the fallback was worse than either of us said

I measured `Notes.from_xbrl` and the concept index behind `StatementLineItem.note` across the five committed fixture directories that yield roles, all of which have `_filing_summary is None` and so take this path:

| fixture | notes (main) | notes (this PR) | concept index (main) | concept index (PR) |
|---|---|---|---|---|
| gahc | 4 | **7** | 22 | **92** |
| aapl | 0 | **16** | 0 | **261** |
| tsla | 0 | **12** | 0 | **133** |
| unp  | 0 | **20** | 0 | **318** |
| aeon | 0 | **12** | 0 | **252** |

My review framed this as "your change empties the note bucket." That was only half of it: `_build_from_xbrl_only` required `category == 'note' and type == 'Notes'`, and four of these five filings never had a single role in that bucket, so `StatementLineItem.note` and `get_notes_for_concept()` already returned nothing for them on main. Your rewrite doesn't just restore what #1207 and this PR moved, it makes the fallback work at all. The notes it now produces are the real ones — aapl's 16 are exactly the 10-K's notes, each with its Tables/Policies/Details attached.

## The precision points hold

Re-running the probes from my first review against the new segment-aligned matcher:

```
stem Debt                              -> DebtorNotesDetails                          rejected   (was FOLLOWS)
stem ConvertibleNote                   -> ConvertibleNoteholderRightsDetails          rejected   (was FOLLOWS)
stem NotesPayableRelatedParty          -> NotesPayableRelatedPartyDetails              FOLLOWS
stem ConvertiblePromissoryNotesPayable -> ConvertiblePromissoryNotePayableScheduleOf…  FOLLOWS
empty stem key ()                      -> anything                                     rejected
Inventory -> InventoriesDetails / Liability -> LiabilitiesDetails                      rejected, documented
```

Genuine members still follow, the two strangers no longer do, and `_role_family_key('')` is now `()` which `_is_family_stem_of` rejects on the length test. The `-ies` limitation is stated in `_segments_match` with a note to widen it when a fixture shows the shape, which is the right call.

I also checked the `.group(0)` on the suffix match in `_build_from_xbrl_only`, which would be an `AttributeError` on a `None`. It's safe by construction: `_role_family_members` only returns definitions the same regex matched, and the loop skips anything not in that set.

The new tests are the ones I asked for, including `get_notes_for_concept("us-gaap_ConvertibleNotesPayableCurrent", xbrl)` resolving to the convertible note's role — that's the end-to-end assertion the old test was missing. `tests/xbrl` plus the regression file is 934 passed; all three CI-only gates pass locally.

## One follow-up, not blocking

A filing that spells *some* roles with a leading section marker and others without splits a family, because the segment prefix is anchored at position 0. `unp` does exactly this:

```
DisclosureDebt          (stem)      ('disclosure','debt')
DisclosureDebtDetails1  MEMBER      ('disclosure','debt','details','1')
DebtDetails2            stem  <--   ('debt','details','2')
DebtDetails6            stem  <--   ('debt','details','6')

Leases                  (stem)      ('leases',)
LeasesDetails           MEMBER
DisclosureLeasesTables  stem  <--   ('disclosure','leases','tables')
```

So `unp` gets two spurious top-level notes (`Debt Details6`, `Disclosure Leases Tables`) and `Disclosure Debt` loses two Details. Strictly better than main's zero notes, so it isn't a regression — just the next layer.

Dropping a leading section marker from the key looks promising:

```python
    segments = [segment.lower() for segment in _CAMEL_SEGMENT_RE.findall(definition)]
    if segments and segments[0] in ('disclosure', 'disclosures', 'statement', 'statements'):
        segments = segments[1:]
    return tuple(segments)
```

On the corpus that gives unp 20 -> 18 notes with the two spurious entries absorbed and `Disclosure Debt` details 5 -> 7, every other fixture unchanged, and 934 still passing. I'd treat that as promising rather than proven — as I noted last time the effective corpus here is five filings, so it isn't much of a test of a normalisation rule, and a leading-marker allowance could plausibly merge two genuinely separate families in a filing none of us has looked at. Entirely fine to leave for a separate issue; I'm happy to file one.

Good revision — the rewrite of the fallback builder is a bigger improvement than the issue asked for, and it's well tested.

PR #1330, opened 2026-09-17
fix: include combined debt concepts in balance sheet fallback mappings

wittling:
## Problem

`balance_sheet()` silently drops debt for companies that report it under combined us-gaap concepts. 

The balance-sheet fallback in `mappings_loader.py` (`_FALLBACK_CONCEPT_MAPPINGS`) only classified `LongTermDebt`, `LongTermDebtCurrent`, and `LongTermDebtNoncurrent` as appearing on the balance sheet. When a filer uses a combined or short-term variant instead — e.g. `LongTermDebtAndCapitalLeaseObligations` (CSX, WSO, Hilton), `DebtCurrent` (PepsiCo), or `ShortTermBorrowings` — the concept resolves to None from `get_all_statements_for_concept()`, and `_fact_belongs_to_statement()` drops the fact before the orphan-facts collector (`_add_orphan_facts`) can surface it. Net effect: total debt is missing entirely from the returned balance sheet for such filers.

## Before (v5.56.0)

```
CSX  -> []
HLT  -> []
WSO  -> []
PEP  -> [LongTermDebtNoncurrent 42,321,000,000]   # DebtCurrent (6.86B) missing
```

## After

```
CSX  -> [LongTermDebtNoncurrent 18,165,000,000, LongTermDebtAndCapitalLeaseObligationsCurrent 708,000,000]
HLT  -> [LongTermDebtNoncurrent 12,338,000,000, LongTermDebtAndCapitalLeaseObligationsCurrent 25,000,000]
WSO  -> [LongTermDebtAndCapitalLeaseObligations 361,635,000, LongTermDebtAndCapitalLeaseObligationsCurrent 117,153,000]
PEP  -> [LongTermDebtNoncurrent 42,321,000,000, DebtCurrent 6,861,000,000]
```

## Fix

Add the common us-gaap combined/current debt concepts to `_FALLBACK_CONCEPT_MAPPINGS` under the BalanceSheet group: the `LongTermDebtAndCapitalLeaseObligations*` variants, `DebtCurrent`, current portions, `ShortTermBorrowings`, and `ShortTermDebt`.


dgunning (commented):
Thanks — this is a real gap and the diagnosis is right. I reproduced your numbers against live data before and after the change:

```
PEP  main: Long-term Debt, Excluding Current Maturities 42,321,000,000
     PR:   + Debt, Current 6,861,000,000
WSO  main: (no debt rows at all)
     PR:   Long-term Debt and Lease Obligation 361,635,000
           Long-term Debt and Lease Obligation, Current 117,153,000
```

The mechanism is sound: `_FALLBACK_CONCEPT_MAPPINGS` is merged in `_build_concept_statement_index` only for concepts not already present from learned data, so this supplements rather than overrides. All nine resolve to `None` before and `BalanceSheet` after. `tests/entity`, `test_issue_644_debt_disambiguation.py` and `test_issue_438_regression.py` all still pass (165 tests).

Three things before I merge it.

## 1. Three of the nine names do not look like real filed tags

Checked against `data/xbrl-mappings/gaap_taxonomy_mapping.csv`, which lists tags actually observed in filings, counting each as a *source* tag:

| concept | observed as a filed tag |
|---|---|
| `LongTermDebtAndCapitalLeaseObligations` | yes |
| `LongTermDebtAndCapitalLeaseObligationsCurrent` | yes |
| `LongTermDebtAndCapitalLeaseObligationsIncludingCurrentMaturities` | yes |
| `DebtCurrent` | yes |
| `ShortTermBorrowings` | yes (deprecated 2018) |
| `ShortTermDebt` | canonical target name, not a source tag |
| `LongTermDebtAndCapitalLeaseObligationsNoncurrent` | no occurrences |
| `CurrentPortionOfLongTermDebt` | no occurrences |
| `CurrentPortionOfCurrentContractualObligations` | no occurrences anywhere in the repo |

`CurrentPortionOfLongTermDebt` in particular is this codebase's own canonical output label (`edgar/xbrl/standardization/section_membership.json:28`), never a source concept; the us-gaap element for that idea is `LongTermDebtCurrent`, which is already in the table. The other two I could not corroborate at all.

Corroboration from the other direction: the XBRL-side mapping (`edgar/xbrl/standardization/gaap_mappings.json`) already maps all six genuine concepts to `BalanceSheet` and is missing exactly these same three. That both supports the six and suggests the three are not real elements.

Caveat: that CSV reflects a corpus of observed filings rather than the complete us-gaap dictionary, so absence is not proof. If you took any of the three from a specific filing, say which and I will take them as-is. Otherwise please drop them — they are inert, but they are noise in a hand-maintained table.

## 2. A changelog fragment

Please add `changelog.d/<issue-or-slug>.fixed.md` rather than editing `CHANGELOG.md` (see `changelog.d/README.md`). A bold headline plus a sentence or two with one measured value, 500 characters max. The PepsiCo figure is a good one to use.

## 3. A regression test

The project's definition of done asks for one ground-truth assertion from a real filing. You already have four companies with hand-checked values, so this is nearly free — assert that `DebtCurrent` and `LongTermDebtAndCapitalLeaseObligationsCurrent` resolve to `BalanceSheet` through `get_concept_statements()`, which needs no network, and optionally one networked check that PepsiCo's balance sheet carries the 6,861,000,000 current-debt row. `tests/issues/regression/` is the place.

I have approved the workflow runs so you get real CI results on this head. Happy to merge once the fragment and test are in and the three names are resolved.

PR #1332, opened 2026-09-18
deprecate EDGAR_ACCESS_MODE and the NORMAL/CAUTION/CRAWL access modes (#1326)

### LINKED ISSUE #1326: EDGAR_ACCESS_MODE is documented with a timeout/connection/retry table, but nothing in the package reads it
## Issue Type
- [x] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.58.0
**Python Version:** 3.14.3

## Bug Description
`docs/configuration.md` documents `EDGAR_ACCESS_MODE` (`NORMAL`, `CAUTION`, `CRAWL`) with a table of HTTP timeouts, connection limits and retry counts per mode, and recommends `CAUTION` or `CRAWL` for being a good SEC citizen. I expected setting `EDGAR_ACCESS_MODE=CRAWL` to change the client's timeout, connection count or retries.

**What actually happened:**
Nothing changes. The variable is read — `edgar/settings.py` (lines 90-99) builds `edgar_mode` from it, and `edgar/core.py` and `edgar/__init__.py` re-export `edgar_mode`, `CAUTION`, `CRAWL` and `NORMAL` — but no module in the package reads `edgar_mode` or `edgar_access_mode`, and nothing reads the mode's fields. `HTTP_MGR` in `edgar/httpclient.py` is built from `EDGAR_HTTP_TIMEOUT` and `EDGAR_RATE_LIMIT_PER_SEC` alone. So the documented setting takes effect silently: it looks configured and does nothing.

**Error message (if any):**
```
none — that is the problem; the setting fails silently
```

## Reproduction
```bash
grep -rn "EDGAR_ACCESS_MODE\|edgar_mode\b\|edgar_access_mode" edgar --include=*.py
```
At 5.58.0 this lists only `edgar/settings.py` (the read and the three assignments), `edgar/core.py` (two re-export lines) and `edgar/__init__.py` (one import line). `docs/configuration.md` names `EDGAR_ACCESS_MODE` seventeen times (lines 45-69, 316, 339, 391, 616-647, 673, 695, 731, 768 and 786).

## Additional Context
Two ways to close this and I am happy to send a PR for either:
1. Wire it: have `get_http_mgr` read `edgar_mode.http_timeout` / `max_connections` / `retries` (or whatever the mode objects carry) so the table becomes true.
2. Drop it: remove the `EDGAR_ACCESS_MODE` section from `docs/configuration.md` and deprecate the exports, pointing readers at `EDGAR_RATE_LIMIT_PER_SEC` and `EDGAR_HTTP_TIMEOUT`, which do work.

Which do you prefer? A separate docs PR (coming) fixes three names in the same docs that no longer exist (`set_rate_limit`, `enable_local_storage`) and the submissions cache TTL; this one is kept apart because it is a design choice rather than a typo.

0ywfe:
Fixes #1326.

Implements the scope agreed in [your comment](https://github.com/dgunning/edgartools/issues/1326#issuecomment-5706744702): drop the modes rather than wire them up, deprecate with a pointer at the two settings that work, and take `EdgarSettings.limits` with them.

## What was wrong

`NORMAL` / `CAUTION` / `CRAWL` advertise a timeout, connection-limit and retry policy that was never wired into the HTTP client. Nothing in the package reads `edgar_mode`, `.retries` or `EdgarSettings.limits`, so all three modes have always behaved identically and `EDGAR_ACCESS_MODE` has never changed how edgartools talks to SEC.

The docs are the part that made it costly. `configuration.md` presented the policy as a table of concrete timeouts and connection counts, so someone picking `CRAWL` for a bulk job had every reason to think they were throttling themselves.

## What this does

1. Docs: the `EDGAR_ACCESS_MODE` section and the per-mode table are replaced with a deprecation notice pointing at `EDGAR_RATE_LIMIT_PER_SEC` and `EDGAR_HTTP_TIMEOUT`, plus a worked example of the settings that replace `CAUTION`/`CRAWL`.
2. `DeprecationWarning` on `EDGAR_ACCESS_MODE` being set, and on access to `NORMAL`, `CAUTION`, `CRAWL`, `edgar_mode` and `edgar_access_mode` from any of the three import paths. Exports keep working; removal lands in 6.0.
3. Tests: `tests/issues/regression/test_issue_1326_access_mode_deprecation.py`.
4. `EdgarSettings.limits` dropped, not repaired.

## Notes for review

Identity had to be preserved. `test_07lk121_settings_extraction.py` asserts the re-exports are the *same objects* across `edgar` / `edgar.core` / `edgar.settings`. A `__getattr__` that rebuilt an `EdgarSettings` per access would pass an equality check and silently break `mode is NORMAL`. The modes stay module-level singletons built once, and `__getattr__` only routes to them. That file still passes unmodified.

`import edgar` had to stay silent. The modes were eager imports in `edgar/__init__.py` and `edgar/core.py`. Left that way, the warning would fire once for every user on every import, naming the re-export site rather than the code reaching for a mode, which trains people to filter a warning instead of acting on it. Both re-exports are now lazy, and a test asserts a plain `import edgar` under `-W error::DeprecationWarning` exits 0.

`edgar/core.py` already had a `__getattr__`, installed via `edgar._compat.deprecated_alias`. A second assignment would have silently replaced the `TooManyRequestsException` alias, so I chained through that helper's existing `__module_getattr__` fall-through. It adds a stack frame, so that path warns at `stacklevel=4`; at the default it blamed `edgar/_compat.py:59` instead of the caller. Verified for all three paths and locked by a test, since a deprecation that points at library internals tells the user nothing.

`EdgarSettings.limits` is as you described. It ignored `self.max_connections` and returned the module default, so all three modes reported `max_connections=10` despite declaring 10/5/2. Dropped with the rest. The module-level `limits` object is a different name with real re-export coverage and is untouched.

The import surface is unchanged, and that took one more fix than I expected. `test_public_api_surface.py` classifies the top-level namespace by `dir()`. Making the re-exports lazy removed `edgar_mode` from `dir(edgar)` while leaving it perfectly readable, so its `INTERNAL` entry started reading as a dead one. `edgar/__init__.py` now carries a `__dir__` covering exactly the four names the eager import used to place there, on the view that a name you can read but cannot see is a worse surface than either option.

`edgar_access_mode` is deliberately excluded from that four. It has never been reachable from `edgar` itself, only from `edgar.core` and `edgar.settings`, and my first pass served it there, which would have widened the public surface in a change meant to narrow it. Now pinned by a test asserting the top-level namespace exposes exactly what it did before.

One pre-existing test leak surfaced on the way. `tests/display/test_core.py::test_settings` assigns `edgar.edgar_mode = CRAWL` and never restores it. A real module global shadows a PEP 562 `__getattr__`, so that leak silently suppressed the new deprecation for every test running after it, and only in a full-suite run, not when the new tests run alone. Restored in a `finally`, and the new tests clear any shadowing override in an autouse fixture so they no longer depend on ordering. The shadowing itself is correct behaviour (if you assign it, you own it) and is now pinned by two tests rather than left as folklore.

## Slightly beyond the stated scope

Your item 1 was the section and the table. I also removed 11 `export EDGAR_ACCESS_MODE=...` lines from examples elsewhere in `configuration.md` (Docker, CI, production-config blocks), the diagnostic snippet that printed the access mode, and the one in `examples/scripts/advanced/enterprise_config.py`. Leaving them would have kept the docs recommending a variable that now emits a `DeprecationWarning`. The diagnostic snippet prints the rate limit and timeout instead, and I checked the defaults it quotes against the code (`EDGAR_RATE_LIMIT_PER_SEC` default `9` in `httpclient.py:202`, `EDGAR_HTTP_TIMEOUT` default `30.0` in `httpclient.py:187`).

I also deprecated `edgar_access_mode`, the parsed string, alongside `edgar_mode`. Your list didn't name it explicitly. Leaving it live while everything around it warns seemed incoherent, but say the word and I'll pull it back out.

## One observation, not acted on

`settings.default_http_timeout = 12` looks like a third dead export of the same family. It appears only in `settings.__all__` and the `edgar.core` re-export, and nothing reads it. The timeout that actually applies is `EDGAR_HTTP_TIMEOUT`, defaulting to `30.0`, which is a different number from the one `default_http_timeout` advertises. Out of scope here, but it may belong on the same 6.0 list.

It caught me out too: my first draft of the diagnostic snippet quoted `12` as the `EDGAR_HTTP_TIMEOUT` default, which is exactly the class of error this PR is about. Corrected before pushing.

## Verification

- `tests/issues/regression/test_issue_1326_access_mode_deprecation.py`, full new suite passing
- `test_07lk121_settings_extraction.py`, 9 passed, unmodified
- `ruff check` on every touched file, no new findings against `main` (`edgar/settings.py` 0, `edgar/core.py` and `edgar/__init__.py` unchanged at their pre-existing counts)
- Changelog fragment at `changelog.d/1326.deprecated.md`


dgunning (changes requested):
This is careful work, and the notes-for-review section did most of my job for me. I verified the mechanism in a worktree rather than taking it on trust, and it holds on every point:

```
plain `import edgar` under -W error::DeprecationWarning   exit 0
all six access paths warn (edgar / edgar.core / edgar.settings)
edgar.NORMAL is edgar.settings.NORMAL                     True
edgar.core.CRAWL is edgar.settings.CRAWL                  True
dir(edgar) -> CAUTION, CRAWL, NORMAL, edgar_mode          (edgar_access_mode correctly absent)
EDGAR_ACCESS_MODE set -> warns, message names both replacements
EdgarSettings.limits                                      gone
```

The `stacklevel` arithmetic is right on all three paths — each blames the calling line, including the chained route through `deprecated_alias`. Docs mentions went from 17 to 2, both inside the new notice. The fragment is 403 characters. 228 tests pass across the four affected files, including `test_07lk121_settings_extraction.py` unmodified.

Answering your two questions: yes, deprecating `edgar_access_mode` alongside `edgar_mode` is right, leaving it live would have been incoherent. And you were right to keep it out of `dir(edgar)` — widening the surface in a change meant to narrow it would have been the wrong trade. The `__dir__` reasoning is correct too: a name you can read but cannot see is the worst of the three options.

The `EdgarSettings.limits` removal was my instruction, so that one is on me rather than you. I considered whether it deserved a warning property until 6.0 like everything else in the diff, and I am content with the hard removal: nothing reads it and it returned a wrong value for every mode.

## One thing to fix before I merge

`_run()` in the new test file inherits the ambient environment. `subprocess.run` is called with no `env`, so an inherited `EDGAR_ACCESS_MODE` reaches the child, your new import-time warning fires, and `-W error` makes the child exit non-zero:

```
(unset)                      36 passed
EDGAR_ACCESS_MODE=CRAWL      2 failed, 34 passed
                             TestImportingEdgarIsSilent::test_plain_import_emits_no_deprecation
                             TestImportingEdgarIsSilent::test_importing_edgar_core_emits_no_deprecation
```

So the suite goes red for exactly the people this deprecation is aimed at — anyone who already set the variable. Pass `env={**os.environ}` with `EDGAR_ACCESS_MODE` popped.

While you are in there: `tests/display/test_core.py::test_settings` has the same environment dependence (`assert edgar.edgar_mode.max_connections == 10` is false under `CRAWL`). Pre-existing, but you rewrote that body for leak-safety, so it is a natural moment to close it — `monkeypatch.delenv` plus a reload, or assert against `edgar.settings.NORMAL` directly.

## Two documentation gaps

The replacement table at `docs/configuration.md:57` links `EDGAR_HTTP_TIMEOUT` to `[below](#edgar_http_timeout)`, but no such heading exists in the file, and the variable is also missing from the environment-variable reference table around line 505. It is now recommended in three places and documented with a heading in none, which is a smaller version of the problem this PR exists to fix. Please add the section the link promises.

`docs/upgrade/6.0.md` now contradicts this change: line 569 presents the modes as live settings that merely moved modules, and line 590 reads `from edgar import set_identity, CAUTION   # unaffected, now and in 6.0`. Not your file, but this change creates the contradiction, so it belongs in this PR.

## Your observation

`settings.default_http_timeout = 12` is indeed a third dead export of the same family, and it advertises a number that is wrong twice over — nothing reads it, and the timeout that applies defaults to `30.0`. I have filed it separately for 6.0; no need to touch it here. I also checked `default_max_connections`, which you left alone: still read by the module-level `limits` object, so leaving it was correct.

The scope you went slightly beyond was the right call. Leaving eleven `export EDGAR_ACCESS_MODE=...` lines in the Docker and CI examples would have had the docs recommending a variable that now warns.

Ping me when the test environment fix and the two doc items are in and I will merge.

0ywfe:
All four are in, on a second commit so the delta is separate from the original.

**The environment leak.** You were right about the mechanism and about who it hits. `_run()` now builds the child's environment explicitly with `EDGAR_ACCESS_MODE` removed, and takes keyword overrides so the env-var test passes it in rather than setting it inside the child. I reproduced your numbers before fixing (unset 36 passed; `EDGAR_ACCESS_MODE=CRAWL` 2 failed, 34 passed) and the affected files now pass 88 both ways.

**`test_core.py::test_settings`.** Closed as you suggested. The first assertion now reads `edgar.settings.NORMAL.max_connections == 10`, which is the mode it actually meant. `edgar.edgar_mode` is whichever mode the variable selected at import, so asserting 10 against it was only ever true when the variable was unset.

**The missing heading.** Added `#### EDGAR_HTTP_TIMEOUT` where the link promised it, plus the summary-table row. I documented it from `httpclient.py` rather than paraphrasing: default `30.0`, with `"none"`, `"unlimited"`, `"0"` and empty all meaning no timeout, and non-positive values routing to the unlimited path because httpx treats a `0.0` read timeout as immediate-timeout. That last one seemed worth stating since `EDGAR_HTTP_TIMEOUT=0` reads like "no timeout" and would otherwise be a trap.

**`docs/upgrade/6.0.md`.** The `# unaffected, now and in 6.0` line was flatly wrong, so I split the example, pointed the settings-move section at the deprecation, and added an "Access modes are deprecated" section giving the two replacements.

Verification: full fast suite 7579 passed. Affected files 88 passed with and without `EDGAR_ACCESS_MODE` set. `ruff` unchanged against `main` (`test_core.py` at its pre-existing 8, the new test file 0). Both `check_regression_provenance.py` and `check_regression_skips.py` green at 387 files.

Thanks for checking the mechanism in a worktree rather than taking the description on trust. The environment dependence is the kind of thing that only shows up if someone actually sets the variable, and I had not thought to test as the user the change is aimed at.

Issue #574, opened 2026-01-02
Dimensions data to include axis, member, and label


Velikolay:
## Feature Category

- [x] New API functionality
- [ ] Performance improvement
- [x] Developer experience improvement
- [ ] Documentation enhancement
- [ ] Tool/utility addition

## Problem Statement
**Is your feature request related to a problem? Please describe.**
To make the most of the dimensions I believe having a clear distinction between these three is vital in statements and xbrl queries.

For example now in the AAPL `income_statement(include_dimensions=True)` dataframe we get
concept: `us-gaap:SalesRevenueNet`
label: `iPhone`
dimension_label: `srt:ProductOrServiceAxis: iPhone`

The proposed structure is more like
concept: `us-gaap:SalesRevenueNet`
label: `Revenue`
dimension_axis: `srt:ProductOrServiceAxis`
dimension_member: `aapl:IPhoneMember`
dimension_label: `iPhone`

This I believe gives the full context of the dimension in a structured form.
* Separates the axis info from the label (better API)
* Adds the `member` which is important for normalisation
* Preserves the main metric label

**Who would benefit from this feature?**
- [ ] Beginner Python users working with SEC filings
- [ ] Financial analysts and researchers
- [x] Advanced developers building financial applications
- [x] Data scientists working with financial datasets

## Proposed Solution
**Describe the solution you'd like**

New structure for dimensional data in statement dataframe, e.g.

concept: `us-gaap:SalesRevenueNet`
label: `Revenue`
dimension_axis: `srt:ProductOrServiceAxis`
dimension_member: `aapl:IPhoneMember`
dimension_label: `iPhone`

Additionally, it would be nice to have the `dimension_label` - e.g. `iPhone` in queries `by_dimension`
```
xbrl.query(include_dimensions=True)
                .by_concept("Revenue")
                .by_dimension("ProductOrServiceAxis")
                .to_dataframe()
```
These already include the `axis` and the `member`, but not the label used in the report.

<img width="239" height="226" alt="Image" src="https://github.com/user-attachments/assets/bdd6512b-9056-4270-a401-09ca86465052" />

**Describe alternatives you've considered**

## Use Case Example
**How would you use this feature?**
```python
# Provide a code example of how you envision using this feature
from edgar import Company

company = Company("AAPL")
# ... example usage of proposed feature
```

## Implementation Considerations

**Complexity Level:**
- [x] Simple (minor API addition)
- [ ] Moderate (new functionality with existing patterns)
[...]

Issue #603, opened 2026-01-16
Dataframe field dimension_member_label has incorrect values


Velikolay:
## Issue Type

- [x] Data Quality Issue (incorrect financial values, missing data)
- [x] XBRL Parsing Issue (statement rendering, concept mapping)  
- [ ] Filing Access Issue (download failures, attachment problems)
- [ ] Performance Issue (slow operations, memory problems)
- [ ] Compatibility Issue (platform/dependency conflicts)
- [ ] Documentation Issue (unclear instructions, missing examples)

## Environment
**EdgarTools Version:** 5.9.1
**Python Version:** 3.13  
**Operating System:** (e.g., macOS 14.0, Windows 11, Ubuntu 22.04)

## Bug Description
**What you expected to happen:**
I would expect the `dimension_member_label` field to contain the member name as shown in the report. Ie for GOOGL they should be something like

Google Search & other
YouTube ads
Google Network
Google advertising
Google subscriptions, platforms, and devices
Google Services total
Google Cloud
Other Bets

**What actually happened:**
The column values are `Google Services` for some of these

<img width="698" height="238" alt="Image" src="https://github.com/user-attachments/assets/9e7d70e3-494a-4225-920c-cfece2f6831a" />

This is also reproducible when the dimension is queried directly, as well as in statements
```
filing.xbrl().query().by_concept("Revenue").by_dimension("ProductOrServiceAxis")
```

**Error message (if any):**
```
Paste any error messages here
```

## Reproduction
**Company/Ticker:** GOOGL
**Time Period:** All (example 2023)
**Relevant Forms:** 10-K, 10-Q

**Minimal code to reproduce:**
```python
company = Company("GOOGL")  # Change ticker as needed
filing = company.get_filings(form="10-K", year=2023).latest()

# result_df = filing.xbrl().query().by_concept("Revenue").by_dimension("ProductOrServiceAxis")
result_df = filing.xbrl().statements.income_statement().to_dataframe()
result_df
# ... rest of reproduction code
```

## Additional Context
Add any other context about the problem here. Screenshots, related issues, etc.

---
*This issue will be handled using EdgarTools' systematic issue resolution workflow. A reproduction test will be created to verify the fix.*

dgunning:
Thank you for the detailed bug report, Nikolay! I've confirmed this is a **data accuracy issue** and have implemented a fix.

### Issue Confirmed

I reproduced the issue with GOOGL's 2023 10-K:

```python
from edgar import Company

[...]

PR #1335, opened 2026-09-21
fix(xbrl): keep two-axis segment revenue in the DETAILED view

### LINKED ISSUE #1331: get_raw_data: Explicit DETAILED view omits certain reportable-segment revenue observations returned by the default view
# get_raw_data: Explicit DETAILED view omits certain reportable-segment revenue observations returned by the default view

## Issue Type

## Environment

**EdgarTools Version:** 5.58.0, freshly built and installed from commit `e23d04eba952e70310c0f62402f4c2523f9a44bf` into a new virtual environment, with its declared dependencies installed. Also reproduced in the original installed 5.56.0 environment.

**Python Version:** CPython 3.14.4, 64-bit, MSC v.1944 (AMD64).

**Operating System:** Windows 11, build 26200 (`Windows-11-10.0.26200-SP0`).

The 5.58.0 reproduction ran from outside both source checkouts. The imported module path was `C:/dev/issue-reports/edgartools/.work/stable-runtime/Lib/site-packages/edgar/__init__.py`, confirming it used the freshly installed target build.

The same result was then verified after upgrading the active commissioned runtime to 5.58.0 from the same commit, importing `C:/dev/plicit/library/.venv/Lib/site-packages/edgar/__init__.py`. The minimal code block below was extracted from this report and executed verbatim in that upgraded runtime.

## Bug Description

For the same public Notes child Statement, `get_raw_data(view=StatementView.DETAILED)` should retain the two reportable-segment revenue rows and their six annual values that are available from `get_raw_data()`.

The public `get_raw_data` signature accepts `StatementView`, and the enum's public help describes DETAILED as including dimensional breakdowns with relaxed member filtering for complete extraction. The report uses the enum as its primary reproducer; it does not rely on whether the string form is supported by this method.

**What actually happened:**

The default call returns revenue for Compute & Networking and Graphics for FY2026, FY2025 and FY2024. Explicit `StatementView.DETAILED` returns no revenue observations with `us-gaap:StatementBusinessSegmentsAxis`. Other rows remain in the returned Statement. A subsequent default call returns the six values again.

Both reportable-segment rows carry a second dimension: `srt:ConsolidationItemsAxis = us-gaap_OperatingSegmentsMember`. These are distinct from product/end-market breakdowns.

| Call on the same Statement instance | 5.58.0, requested commit | Original installed 5.56.0 |
|---|---:|---:|
| `get_raw_data()` before explicit views | 6 | 6 |
| `get_raw_data(view="detailed")` | 0 | 0 |
| `get_raw_data(view=StatementView.DETAILED)` | 0 | 0 |
| `get_raw_data()` after explicit views | 6 | 6 |

Counts are numeric annual revenue observations on the reportable-business-segment axis, not total Statement row counts.

**Error message (if any):**

No exception or warning is raised. The omission is silent.

## Reproduction

**Company/Ticker:** NVIDIA Corporation / NVDA, CIK 1045810.

**Time Period:** FY2026 ended January 25, 2026; FY2025 ended January 26, 2025; FY2024 ended January 28, 2024.

**Relevant Forms:** 10-K filed February 25, 2026, accession `0001045810-26-000021`.

Set `EDGAR_IDENTITY` to a valid SEC contact identity before running.

```python
import os
from edgar import get_by_accession_number, set_identity
from edgar.xbrl.presentation import StatementView

set_identity(os.environ["EDGAR_IDENTITY"])
filing = get_by_accession_number("0001045810-26-000021")
note = next(n for n in filing.obj().notes if n.title == "Segment Information")
statement = next(
    child for child in note.children
    if child.role_or_type.endswith(
        "/SegmentInformationScheduleofRevenuebyMarketDetails"
    )
)

def segment_revenue_count(rows):
    return sum(
        sum(isinstance(v, (int, float)) and not isinstance(v, bool)
            for v in row.get("values", {}).values())
        for row in rows
        if row.get("concept") == "us-gaap_Revenues"
        and any(d["dimension"] == "us-gaap:StatementBusinessSegmentsAxis"
                for d in row.get("dimension_metadata", []))
    )

print("default before:", segment_revenue_count(statement.get_raw_data()))
print("detailed enum:", segment_revenue_count(
    statement.get_raw_data(view=StatementView.DETAILED)))
print("default after:", segment_revenue_count(statement.get_raw_data()))
```

Actual output on 5.58.0 at the specified commit:

```text
default before: 6
detailed enum: 0
default after: 6
```

Expected: the detailed enum call retains all six observations.

## Additional Context

The following amounts are returned by the default public Statement view and agree with the reportable-segment table in Note 16 of the [filed annual report](https://www.sec.gov/Archives/edgar/data/1045810/000104581026000021/nvda-20260125.htm). USD millions:

| Reportable segment | FY2026 | FY2025 | FY2024 |
|---|---:|---:|---:|
| Compute & Networking | 193,479 | 116,193 | 47,405 |
| Graphics | 22,459 | 14,304 | 13,517 |

The numeric values in the reproduction's raw output are USD, not millions. Exact duration keys are:

- `duration_2025-01-27_2026-01-25`
- `duration_2024-01-29_2025-01-26`
- `duration_2023-01-30_2024-01-28`

The two business members are `nvda_ComputeAndNetworkingSegmentMember` and `nvda_GraphicsSegmentMember`. The selected role is `http://www.nvidia.com/role/SegmentInformationScheduleofRevenuebyMarketDetails` (child index 5 in the tested filing). The reproducer selects by role instead of depending on that index.

**Practical impact:** A consumer requesting detailed rows to gather all numeric business-segment measures can incorrectly conclude that segment revenue is unavailable, even while other segment measures are returned through other Statements. The evidence exists in the filing and in the default view; this is not a disclosure gap.

**Scope of the report:** Confirmed observable public-API difference for this filing and these versions. Internal cause is not established. No assertion is made that every possible Statement must contain all filing facts. The result also occurred with the explicit enum, ruling out string-versus-enum caller ambiguity in this reproduction.

**Public documentation checked at the requested commit:**

- [Detailed dimensional views and accepted view forms](https://github.com/dgunning/edgartools/blob/e23d04eba952e70310c0f62402f4c2523f9a44bf/docs/xbrl/concepts/dimension-handling.md#statement-methods)
- [Statement extraction and detail levels](https://github.com/dgunning/edgartools/blob/e23d04eba952e70310c0f62402f4c2523f9a44bf/docs/guides/extract-statements.md#controlling-detail-level-with-view)
- Bound public help for `Statement.get_raw_data` and public `StatementView` enum help in the installed target build. The documentation examples primarily cover `to_dataframe`; the enum annotation and reproduction establish the supported `get_raw_data` case directly.

---

Dxfory:
## Summary

Fixes #1331.

On NVIDIA's FY2026 10-K (`0001045810-26-000021`), `get_raw_data()` kept the six reportable-segment revenue observations (Compute & Networking $193.479B, Graphics $22.459B). `get_raw_data(view=StatementView.DETAILED)` returned none of them and showed ProductOrService `Compute` / `Networking` instead. Same 16-row statement, swapped content, no warning.

The renderer and `_merge_complementary_rows` are innocent, as @dgunning noted. `get_statement` already had the wrong rows. DETAILED's GH-574 bypass correctly *includes* Data Center's Compute/Networking children. That is what activated `_apply_member_hierarchy`. The hierarchy map keyed every dimensional row by `dimension_metadata[0].member`. Both segment combinations share `OperatingSegmentsMember` on `srt:ConsolidationItemsAxis` (sorted first), so last-wins dropped them the moment any same-axis parent/child pair existed.

Hierarchy now nests only single-axis members and leaves multi-axis combinations in place. DETAILED still surfaces Compute/Networking under Data Center.

## Test plan

- [x] Offline fixture: `tests/fixtures/xbrl/nvda/10k_2026/` — instance + schema + linkbases from the 21 Feb 2026 XBRL zip, recorded 21 Sep 2026 against `main` before the patch.
- [x] `python -m pytest tests/issues/regression/test_issue_1331_detailed_segment_revenue.py tests/issues/regression/test_issue_1224_orphan_member_headings.py -q` — 7 passed (default and DETAILED keep the six filed values; DETAILED still has Compute/Networking; Tesla member nesting still works).
- [x] `python scripts/check_regression_provenance.py` — OK.
- [x] `python scripts/release/assemble_changelog.py --check` — `1331.fixed.md` accepted.

## AI disclosure

Drafted with AI assistance in Cursor (Grok 4.6). The pytest results above were run on this machine, not transcribed from the model. I answer for every line.


dgunning (approved):
Verified and approving. The diagnosis is exactly right, and I confirmed it against the raw fixture rather than taking the description on faith.

## The defect is real and the ground truth is correct

Contexts `c-193` and `c-194` in `nvda-20260125_htm.xml` each carry two axes:

```
c-193  srt:ConsolidationItemsAxis=us-gaap:OperatingSegmentsMember
       us-gaap:StatementBusinessSegmentsAxis=nvda:ComputeAndNetworkingSegmentMember
       2025-01-27 → 2026-01-25   us-gaap:Revenues = 193,479,000,000
c-194  ...=us-gaap:OperatingSegmentsMember + nvda:GraphicsSegmentMember
       same period              us-gaap:Revenues =  22,459,000,000
```

Both share `us-gaap:OperatingSegmentsMember` as `meta[0]`, so the old `member_to_item[meta[0]['member']] = item` collapsed them to one entry, and since `ordered` was built only from `member_to_item.values()`, the loser vanished. Every value asserted in the test matches the filed fact, and `c-235`/`c-238` confirm Compute/Networking are genuinely single-axis `srt:ProductOrServiceAxis` rows — so the second test is guarding the right thing.

## The test actually catches the bug

I reverted `edgar/xbrl/xbrl.py` to `main` and reran, which is the check that matters:

```
FAILED test_default_and_detailed_keep_the_filed_segment_revenue - detailed dropped Compute & Networking
FAILED test_hierarchy_does_not_collapse_two_axis_rows_onto_the_first_member
2 failed, 1 passed
```

Two fail reproducing the reported symptom; `test_detailed_still_includes_data_center_product_children` passes on both sides, which is correct — it exists to catch a regression of the GH #574 behaviour this change could plausibly have broken. The whole file runs offline in 0.11s. All three CI-only gates pass locally (`check_regression_skips`, `check_regression_provenance`, `check_cassettes`).

## Measured blast radius

I instrumented `_apply_member_hierarchy` to assert it is a permutation of its input, and drove every statement in all 28 XBRL fixture directories under both the default and DETAILED views — 13,490 calls, 4,627 of them involving at least one multi-axis row:

| | lossy calls | reorders |
|---|---|---|
| main | **209** | 188 |
| this PR | **7** | 274 |

So it removes 202 of 209 row-loss sites and introduces none. `tests/xbrl` is 900 passed.

I also checked the other `dimension_metadata[0]` sites (`statements.py:1550`, `current_period.py:517` and `:1019`, `facts.py:1509`). Those take the primary dimension to *populate* label/axis/member fields — non-lossy and deliberate per #603/#574 — so this is not a "one rule, two implementations" case. `_apply_member_hierarchy` was the only place where `meta[0]` decided which rows survive.

## One residual case — pre-existing, not a blocker

The 7 remaining losses are a different, narrower bug that this PR inherits rather than causes. `member_to_item` is keyed on the member alone, so the same member filed on *two different axes* still collides. JPM's 2013 10-K does it on `us-gaap_LoansAndLeasesReceivableNetOfDeferredIncome`:

```
dei:LegalEntityAxis:                                    VariableInterestEntityPrimaryBeneficiaryMember
us-gaap:VariableInterestEntitiesByClassificationOfEntityAxis:  (same member)
```

31 rows in, 30 out — the `LegalEntityAxis` row is dropped.

Since the domain hierarchy can't say which axis's row it nests, the safe treatment is to let neither participate:

```python
        member_to_item = {}
        ambiguous = set()
        for item in dim_items:
            meta = item.get('dimension_metadata') or []
            if len(meta) != 1:
                continue
            member_id = meta[0].get('member')
            if not member_id:
                continue
            if member_id in member_to_item:
                # Same member filed on two different axes: the domain hierarchy
                # cannot say which row it nests, so neither participates.
                ambiguous.add(member_id)
                continue
            member_to_item[member_id] = item
        for member_id in ambiguous:
            member_to_item.pop(member_id, None)
```

I tested this on top of your branch: losses go 7 → 0, and `tests/xbrl` plus your regression file stay green at 903 passed. Entirely your call whether to fold it in here or leave it — it predates this PR and I'm happy to take it as a separate issue either way. Nothing blocking; this is good work and the fixture-plus-unit-test pairing is the right shape.

